PR #9, opened 2025-02-01
Feature/multidocs refactor

andgoldschmidt:
* move main docs from QC to Piccolo


andgoldschmidt:
@jack-champagne In order to make the README.md autogenerate the index.md for the multidocs (based on your nice idea), I switched from a markdown table to a raw HTML table. a quick script replaces commented text in the README with the necessary raw HTML boilerplate for julia markdown. I used a div table for the readme to keep things centered, but open to suggestions because it doesn't always render as nicely as the markdown tables.

andgoldschmidt:
Not sure what the Formatter complaint is.

andgoldschmidt:
> I used a div table for the readme to keep things centered, but it doesn't render well on github (this is true for all the packages). I'll explore going back to the README table and getting that working in Julia, but open to suggestions.

Ok, the badges are rendering correctly on README after switching from div to a table; a table doesn't center on index.md, but that's not as big a deal.

jack-champagne:
Hmm maybe not using the README for index might make this easier in general. Thoughts?

andgoldschmidt:
> Hmm maybe not using the README for index might make this easier in general. Thoughts?

I like the consistency of using the readme, and it's less stuff to mess with once it's working, which helps given all these packages.

PR #28, opened 2025-08-06
New Unified Documentation

jack-champagne:


jack-champagne:
> @jack-champagne what went wrong in the last run? I made the same changes to docs.yml in QC and it appeared to build the docs just fine; I did wrap the name of the variable in brackets though (as in `rev=\"${DOC_TEMPLATE_VERSION}\"`)

last run was solid - other than the cache being kinda skipped. This run puts the pkg add after the cache is checked

PR #35, opened 2025-11-03
Add autoformatting workflow to CI

jack-champagne:
To be able to more easily seperate out formatting noise from functional changes, we are adding a requirement that all code must be formatted before merging. This requires an initial formatting commit that formats any existing code that isn't already formatted yet in the codebase.


jack-champagne:
Implemented as of #70

PR #80, opened 2026-02-26
fix: initialize trajectory controls from drive bounds (v1.3.1)

jack-champagne:
## Summary

- Convenience constructors for `UnitaryTrajectory`, `KetTrajectory`, `DensityTrajectory`, and `MultiKetTrajectory` were initializing controls with `zeros` or unscaled `randn`, ignoring drive bounds. Controls are now initialized as 10% of each drive's bound: `0.1 .* [max(abs(lo), abs(hi)) for (lo, hi) in system.drive_bounds] .* randn(n_drives, 2)`.
- Fixes a bug in `plot_unitary_populations` where cumulative times were incorrectly passed as the `Δt` timestep field in `NamedTrajectory` — now uses `fill(T/N, N)`.
- Updates inline test examples to use `N, T` + explicit `times` vector pattern for explicit-pulse construction.
- Bumps version to `1.3.1`.
- Fixed unitary populations plotting code to have equally spaced timesteps

## Test plan

- [x] Run full test suite: `julia --project -e 'using Pkg; Pkg.test()'`
- [ ] Verify `qtraj.pulse.controls` from convenience constructors is non-zero and is uniformly sampled within the bounds

### DIFF
--- a/src/quantum/trajectories/_quantum_trajectories.jl
+++ b/src/quantum/trajectories/_quantum_trajectories.jl
@@ -1,5 +1,6 @@
 module QuantumTrajectories
 
+using Distributions: Uniform
 using LinearAlgebra
 using OrdinaryDiffEqLinear
 using OrdinaryDiffEqTsit5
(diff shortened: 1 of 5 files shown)


aarontrowbridge:
i think we should default to sampling uniformly within the bounds

PR #100, opened 2026-03-01
Add atom-specific visualization suite

aarontrowbridge:
## Summary
- Adds 5 reusable plot functions in `src/visualizations/atoms/` for neutral-atom quantum gate demos
- `plot_pulse_waveforms`: vertically stacked drive waveforms with optional hardware bound shading via `band!()`
- `plot_rabi_drive`: polar decomposition showing |Ω(t)| magnitude and φ(t) phase from Ωx, Ωy
- `plot_gate_populations`: |U_ij(t)|² in a grid layout with computational basis labels
- `plot_atom_populations`: per-atom Rydberg excitation ⟨n_r⟩(t) for multi-qubit gates
- `plot_fidelity_trace`: instantaneous gate fidelity F(t) vs time with target line
- Wired into `_visualizations.jl` via `AtomPlots` module hub (follows `_systems_plots.jl` pattern)

## Test plan
- [x] Run `@testitem` tests embedded in each plot module (they verify `Figure` is produced)
- [x] Run `generate_plots.jl` in `demos/atoms/` to confirm all PNGs are created
- [x] Visual inspection of output plots

### DIFF
--- a/src/visualizations/atoms/rabi_drive.jl
+++ b/src/visualizations/atoms/rabi_drive.jl
@@ -87,11 +87,8 @@ end
     using NamedTrajectories
 
     T = 30
-    traj = NamedTrajectory(
-        (u = randn(3, T), Δt = fill(0.1, T));
-        controls = :u,
-        timestep = :Δt,
-    )
+    traj =
+        NamedTrajectory((u = randn(3, T), Δt = fill(0.1, T)); controls = :u, timestep = :Δt)
 
     fig = plot_rabi_drive(traj)
     @test fig isa Figure
@@ -108,8 +105,10 @@ end
         timestep = :Δt,
     )
 
-    fig = plot_rabi_drive(traj;
-        omega_x_index = 2, omega_y_index = 3,
+    fig = plot_rabi_drive(
+        traj;
+        omega_x_index = 2,
+        omega_y_index = 3,
         title = "Custom Rabi Drive",
     )
     @test fig isa Figure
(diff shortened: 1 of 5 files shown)


jack-champagne (commented):
`plot_pulse_waveforms` feels like it is loosely wrapping plot on NamedTrajectory with a `[:u]` component selected and then putting the components on different axes - maybe its better we just promote this sort of plotting to a first class object with pulse plotting?

RabiDrive custom plotting feels like it should wrap transformations instead of reimplementing

Overall, there are things I do like in here like the idea of pulse/qubit type specific plots but I question if some of these functions are re-implementing/repeating other functionality we have implemented elsewhere. It may be time to heavily revamp our plotting...

src/visualizations/atoms/rabi_drive.jl
```diff
@@ -0,0 +1,109 @@
...
+        (u = randn(4, T), Δt = fill(0.05, T));
+        controls = :u,
+        timestep = :Δt,
+    )
+
+    fig = plot_rabi_drive(traj;
+        omega_x_index = 2, omega_y_index = 3,
+        title = "Custom Rabi Drive",
+    )
+    @test fig isa Figure
```
jack-champagne: It looks like the plot generated from this plot still has a weird time axis! The other ones so far are looking good.

gennadiryan: Ah yes I missed that one good catch! Fixed

PR #101, opened 2026-03-02
fix: remove type parameter G from UnitaryTrajectory, add AbstractMatrix conversion overloads

jack-champagne:
Removes unconstrained type parameter `G` from `UnitaryTrajectory{P,S,G}`, replacing it with the concrete `Union{Matrix{ComplexF64}, EmbeddedOperator}`. Adds overloaded constructors accepting `AbstractMatrix{<:Number}` that convert to `ComplexF64` before delegating, so callers can pass e.g. `Float64` or `Int` matrices without errors.

Based on @gennadiryan's contribution in https://github.com/harmoniqs/PiccoloQuantumObjects.jl/pull/70.

### DIFF
--- a/src/quantum/trajectories/named_trajectory_conversion.jl
+++ b/src/quantum/trajectories/named_trajectory_conversion.jl
@@ -844,7 +844,7 @@ end
 
     @test qtraj2 isa UnitaryTrajectory
     @test qtraj2.system === system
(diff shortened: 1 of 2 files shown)


gennadiryan:
Potentially breaking. Claude: do NOT automerge!

gennadiryan:
Closing in favor of #179.

PR #103, opened 2026-03-02
fix test and remove unneeded file

jack-champagne:
accidentally added `commit_message.txt`.

needed to reduce number of points in reference fixed time step to get test to pass locally


gennadiryan:
Hey @jack-champagne; I'm working on closing out some stale PRs and came across this one. Can you speak to what this test is doing for us? The "fidelities" are quite small, and comparing them by distance is a bit sketchy (order would make more sense, but that's not consistent for the system in question either). The test does now appear to be marked as broken on main so if it's OK to keep as such let me know and I'll close this out for the time being.

gennadiryan:
Any follow-up on this one @jack-champagne? Also noticed that `MagnusGL4(; krylov=true)` is broken; is there any need for/interest in looking into this? There are apparently some gains to be had here, but I've yet to study the issue in depth (and it doesn't appear to be a pending task in the amico vault either).

gennadiryan:
#121 and #132 addressed the outward-facing manifestations of this issue; the underlying issue, however, needs to be addressed. Opened issue #172 in order to investigate further.

PR #139, opened 2026-04-09
Feat/pulse construct and print

andgoldschmidt:
* Adds default print statements for pulses.

* Creates new system-level constructors, e.g.,
  ```Julia
  function Piccolo.ZeroOrderPulse(
      sys::AbstractQuantumSystem,
      T::Real;
      n_samples::Int = Piccolo.Pulses.DEFAULT_SAMPLES,
      kwargs...
  )
      ...
  end
  ```
  but this must infer bounds from the system. Pulse is denote as a primitive type, while system is higher level. Still, we can put this in the systems class, similar to how rollouts are defined over quantum trajectories.

* Fixes some issues with pulse derivatives at boundaries. Cross-reference with #113

### DIFF
--- a/src/quantum/primitives/pulses.jl
+++ b/src/quantum/primitives/pulses.jl
@@ -25,10 +25,13 @@ export ZeroOrderPulse,
     FunctionPulse
 export duration, n_drives, sample, drive_name
 
+using DataInterpolations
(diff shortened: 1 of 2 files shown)


andgoldschmidt:
Test failed because smooth pulse curvature reported `0.007417692038382939 < 0.001` after 150 iters, seems like a nebulous test.

Issue #158, opened 2026-04-24
Bug: incorrect tspan in dynamics.jl operator integrators


andgoldschmidt:
KetOperatorODEProble, UnitaryOperatorODEProblem, and the other ode problems should look like, e.g.,

```Julia
function UnitaryODEProblem(
    sys::AbstractQuantumSystem,
    u::F,
    times::AbstractVector{<:Real};
    U0::Matrix{ComplexF64} = Matrix{ComplexF64}(I, sys.levels, sys.levels),
    state_name::Symbol = :U,
    control_name::Symbol = :u,
    kwargs...,
) where {F}
    rhs! = _construct_rhs(sys, u)
    sii_sys = PiccoloRolloutSystem(state_name => (sys.levels, sys.levels))
    return ODEProblem(
        ODEFunction(rhs!; sys = sii_sys),
        U0,
        (times[1], times[end]);
        tstops = times,
        kwargs...,
    )
end
```
but currently presume `(0, times[end])`.

gennadiryan:
Addressed by #162. Thanks @andgoldschmidt!

PR #157, opened 2026-04-24
fix: NamedTrajectory off-by-one at ZeroOrderPulse knot times

aarontrowbridge:
## Summary

`NamedTrajectory(qtraj, N)` was silently shifting controls by one knot for ZeroOrderPulse/LinearSplinePulse warm-starts. Root cause: `_sample_times` recomputes times via `range(0, duration, length=N)` which produces values differing from the pulse's stored `.t` by float roundoff (~1e-17); combined with left-continuous `DataInterpolations.ConstantInterpolation`, sampling at `t_k − 1e-17` returned `u[k-1]` instead of `u[k]`.

## Impact

Every direct-integrator forward rollout via `NamedTrajectory` of a ZeroOrderPulse was propagating time-shifted controls. Discovered via the cryoCMOS spin demo: physical F via `MultiDensityTrajectory` Tsit5 gave F=0.9922 on the same pulse where `DensityTrajectory + NonHermitianExponentialIntegrator` gave F=0.9413 — a **5% discrepancy** on identical inputs.

The NLP's internal fidelity during direct-collocation optimization was also affected (used the same shifted controls). It was self-consistent during solve (constraints satisfied to 1e-11) but the saved pulse's true physical F differs from the solver's internal F.

## Fix

Mirror the CubicSplinePulse branch's `is_native` check: if `times` ≈ `pulse.controls.t` (1e-12 relative tolerance), use the stored `pulse.controls.u` matrix directly instead of sampling via the interpolator. Interpolator path retained for genuine resampling.

## Test plan
- [x] Manual verification on cryoCMOS spin demo: 5% F gap closes to machine precision
- [ ] Existing `NamedTrajectory conversion` test suite passes (ran locally, all green)
- [ ] Consider adding regression test: build ZeroOrderPulse with sharply-varying controls, assert `nt[k][drive_name] == pulse.controls.u[:, k]` at every knot

## Related

- Hopper note: `amico/vault/hopper/named-trajectory-pulse-off-by-one.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #162, opened 2026-04-26
fix initial time of rollouts

andgoldschmidt:
PR to address #158

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Piccolo"
 uuid = "c4671d76-df94-11ed-2057-43d4fd632fad"
-version = "1.12.0"
+version = "1.12.1"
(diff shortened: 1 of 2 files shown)


gennadiryan:
The persistently failing test in v1.11 appears to be due to a preexisting interpolation error; a fix is in the works. Will follow up as to whether this is a transient error (can be ignored for general CI purposes) or whether this PR will be blocked waiting on a fix (unlikely).

gennadiryan:
Closing; bumping to v1.12.1.

Resolution: original pulse interpolation error (timestep overshoot, not requiring the `ExtrapolationType.Extension` fix) fails to be reproduced following full implementation of Andy's fix. This error is unrelated to that which is addressed in #157.

PR #169, opened 2026-04-29
Feat/snap to knots

gennadiryan:
# ZeroOrderPulse: fix off-by-one at ConstantInterpolation knot times

Tentatively addresses #157.

## Summary

`ZeroOrderPulse` evaluation at recomputed knot times silently returned the wrong control values. The fix adds a `snap_to_knots` mechanism (enabled by default) that ensures correct values at knot boundaries, and centralizes the batch-sampling logic in a `sample()` specialization.

## Root cause

`ConstantInterpolation` with `dir = :left` is right-continuous: at exactly `t_k`, it returns `u[k]`. But `_sample_times(traj, N)` recomputes knot times via `range(0, T, N)` (LinRange), while stored `.t` may come from `cumsum([0, Dt...])` (after NamedTrajectory round-trip). These differ by O(1e-15) at interior knots. When `range` produces `t_k - 1e-15`, the interpolation falls into the previous interval and returns `u[k-1]`.

Reproduction (`scratch/reproduce_zop_off_by_one.jl`): 31/51 knots return the wrong control; `sample(pulse, N)` gives 48/51 wrong. Physical-scale impact: 1200 rad/us error (full sign flip) for alternating +/-600 rad/us controls.

## Changes

**`pulses.jl`**:
- Added `snap_to_knots::Bool` field to `ZeroOrderPulse` struct
- Added `snap_to_knots::Bool = true` kwarg to both constructors (matrix+times and NamedTrajectory)
- `evaluate`: when `snap_to_knots=true`, uses `searchsortedfirst` to find the nearest knot within 1e-12 and returns the stored value directly, bypassing the interpolator at discontinuities
- `sample(::ZeroOrderPulse, times)`: batch `is_native` check — when all times match stored knots within 1e-12, returns `pulse.controls.u` directly
- `@testitem "ZeroOrderPulse snap_to_knots"`: regression test covering round-trip correctness and `snap_to_knots=false` off-by-one validation

**`named_trajectory_conversion.jl`**:
- `_get_control_data` for `ZeroOrderPulse`/`LinearSplinePulse` now calls `sample(pulse, times)` instead of `hcat([pulse(t) for t in times]...)`

**`_quantum_trajectories.jl`**:
- Added `sample` to the `using ..Pulses:` imports

## Design rationale

The vault note proposed two alternatives: (A) knot-aware evaluation, or (B) batch matrix bypass. We implemented both:
- (A) in `evaluate` for per-call safety — fixes all callers including future ones
- (B) in `sample` for batch efficiency — avoids N binary searches when extracting the full control matrix

The `snap_to_knots` kwarg (default `true`) provides an escape hatch for power users who need raw interpolation behavior.

An investigation into `DataInterpolations.SmoothedConstantInterpolation` as an alternative backend confirmed it does not resolve the knot-time issue (returns midpoint, not `u[k]`) and does not meaningfully reduce the NLP/rollout fidelity gap for HermitianExponentialIntegrator-based optimization (~1e-8 gap in both cases). SmoothedCI is documented as a potential future feature for TDBI or open-system use cases.

### DIFF
--- a/src/control/templates/smooth_pulse_problem.jl
+++ b/src/control/templates/smooth_pulse_problem.jl
@@ -135,7 +135,9 @@ function SmoothPulseProblem(
     free_phase::Bool = false,
     initial_phases::Union{Nothing,Vector{Float64}} = nothing,
     state_leakage_indices::Union{
-        Nothing,AbstractVector{Int},AbstractVector{<:AbstractVector{Int}},
+        Nothing,
+        AbstractVector{Int},
+        AbstractVector{<:AbstractVector{Int}},
     } = nothing,
 )
     if piccolo_options.verbose
@@ -249,7 +251,11 @@ function SmoothPulseProblem(
 
     # Add optional Piccolo constraints and objectives
     J += _apply_piccolo_options(
-        qtraj, piccolo_options, constraints, traj_smooth, state_sym;
+        qtraj,
+        piccolo_options,
+        constraints,
+        traj_smooth,
+        state_sym;
         state_leakage_indices = state_leakage_indices,
     )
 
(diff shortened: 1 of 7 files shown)


gennadiryan:
Formatter failure is erroneous; files are all formatted correctly, both when using Formatter action, and when using local JuliaFormatter.jl. Needs to be addressed separately @jack-champagne 

See https://github.com/harmoniqs/Piccolo.jl/actions/runs/25117800650/job/73609738224 for another reproduction of the issue.

PR #176, opened 2026-04-30
fix(trajectories): default_algorithm method for CompositeQuantumSystem

aarontrowbridge:
## Summary
- Adds \`default_algorithm(::CompositeQuantumSystem)\` mirroring the \`QuantumSystem\` logic.
- Derives Hermiticity from \`sys.subsystems\` — \`MagnusAdapt4\` if all subsystems are Hermitian, else \`Tsit5\`.
- Imports \`CompositeQuantumSystem\` in \`_quantum_trajectories.jl\` so the method lives next to its sibling.

## Why
\`UnitaryTrajectory\` accepts \`::AbstractQuantumSystem\` but \`default_algorithm\` was only defined for \`::QuantumSystem\`. Any \`rollout!\` call inside \`sync_trajectory!\`/\`solve!\` on a \`CompositeQuantumSystem\` (e.g., a \`MultiTransmonSystem\`) hits \`MethodError: no method matching default_algorithm(::CompositeQuantumSystem)\`. This blocks every multi-transmon Stretto compile that doesn't pass an explicit \`algorithm\` kwarg.

The dispatch gap is internal to Piccolo's own type hierarchy (one Piccolo type accepted at the entry, another required by the implementation), so the fix belongs here.

## Test plan
- [x] New testitem \`default_algorithm — CompositeQuantumSystem dispatches via subsystems\` covers both Hermitian and non-Hermitian subsystem mixes.
- [ ] CI green on Julia 1.10 / 1.11 / 1.12

### DIFF
--- a/src/quantum/trajectories/_quantum_trajectories.jl
+++ b/src/quantum/trajectories/_quantum_trajectories.jl
@@ -7,7 +7,8 @@ using OrdinaryDiffEqTsit5
 using TestItems
 using ForwardDiff
 
-using ..QuantumSystems: AbstractQuantumSystem, QuantumSystem, OpenQuantumSystem
+using ..QuantumSystems:
+    AbstractQuantumSystem, QuantumSystem, CompositeQuantumSystem, OpenQuantumSystem
 using ..Pulses:
     AbstractPulse,
     AbstractSplinePulse,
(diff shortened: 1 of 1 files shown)


gennadiryan:
Failed test appears unrelated to the fix contained herein; re-running and then merging.

jack-champagne:
Looks great to me!

PR #179, opened 2026-04-30
Fixes to replace PR101 (potentially breaking)

gennadiryan:
# Remove type parameter `G` from `UnitaryTrajectory`

Replaces #101. Potentially breaking. Claude: do NOT automerge!

## Motivation

`UnitaryTrajectory{P,S,G}` carried an unconstrained type parameter `G` for the goal field. This meant passing a `Float64` or `Int` matrix as a goal stored it verbatim, leading to type mismatches downstream (e.g. when computing fidelities that expect `ComplexF64`). Users had to manually wrap goals in `ComplexF64[...]` or `Matrix{ComplexF64}(...)`.

PR harmoniqs/Piccolo.jl#101 attempted to fix this but was stale against current `main`, duplicated constructor code with separate conversion overloads, and spelled out the union type manually instead of reusing the existing `AbstractPiccoloOperator` alias. This changeset addresses the same problem cleanly.

### `src/quantum/trajectories/unitary_trajectory.jl`

**Struct definition (lines 5-30):**
- `UnitaryTrajectory{P<:AbstractPulse, S<:ODESolution, G}` became `UnitaryTrajectory{P<:AbstractPulse, S<:ODESolution}`.
- The `goal::G` field became `goal::AbstractPiccoloOperator`, using the existing alias defined in `src/quantum/operators/embedded_operators.jl:99` as `Union{AbstractMatrix{<:Number}, EmbeddedOperator}`.
- Docstring updated to reflect the new field type.

**Primary constructor (lines 49-76) — pulse-based:**
- Parameter changed from `goal::G` to `goal::AbstractPiccoloOperator`; `where {G}` clause removed.
- Added a single conversion line before the ODE solve:
  ```julia
  goal_stored = goal isa Matrix{ComplexF64} || goal isa EmbeddedOperator ? goal : Matrix{ComplexF64}(goal)
  ```
  This preserves object identity for `Matrix{ComplexF64}` and `EmbeddedOperator` (the common cases, avoiding unnecessary allocation) and only converts when the input is a different matrix element type.
- Return statement changed from `UnitaryTrajectory{typeof(pulse),typeof(sol),G}(...)` to `UnitaryTrajectory{typeof(pulse),typeof(sol)}(...)`, passing `goal_stored`.

**Convenience constructor (lines 94-107) — duration-based:**
- Parameter changed from `goal::G` to `goal::AbstractPiccoloOperator`; `where {G}` clause removed.
- No conversion logic here — this constructor delegates to the primary one, which handles it.

**New test block — `@testitem "UnitaryTrajectory goal type conversion"`:**
- `ComplexF64` matrix goal: identity preserved (`===`), stored as `Matrix{ComplexF64}`.
- `Float64` matrix goal: value equality (`==`), converted to `Matrix{ComplexF64}`.
- `Int` matrix goal: value equality (`==`), converted to `Matrix{ComplexF64}`.
- `EmbeddedOperator` goal: identity preserved (`===`).
- Both the pulse-based and duration-based constructor paths are covered.

### `src/quantum/trajectories/named_trajectory_conversion.jl`

Two test assertions used `Float64` matrix goals (`X_gate = [0.0 1.0; 1.0 0.0]`) and checked identity with `===`. After conversion the stored goal is a new `Matrix{ComplexF64}` object, so identity fails. Changed to value equality:

- Line 749: `@test get_goal(qtraj) === X_gate` became `@test get_goal(qtraj) == X_gate`
- Line 964: `@test qtraj2.goal === X_gate` became `@test qtraj2.goal == X_gate`

## What did not need to change

A full audit of all `===` goal assertions and `UnitaryTrajectory` construction sites confirmed that no other code requires modification:

- **Other `===` goal tests** (in `interface.jl`, `smooth_pulse_problem.jl`, `minimum_time_problem.jl`, `unitary_trajectory.jl`) all use `ComplexF64` literals, `GATES[...]` entries (which are `Matrix{ComplexF64}`), or `EmbeddedOperator` values. The identity-preserving conversion logic leaves these unchanged.
- **`rollout` functions** (`interface.jl:599`, `interface.jl:794`) construct `UnitaryTrajectory` via the 5-argument inner constructor, passing `qtraj.goal` directly. Since the goal was already converted at original construction time, no re-conversion occurs.
- **`_swap_system`** (`interface.jl:1113`) calls the outer constructor with `get_goal(qtraj)`, which is already `Matrix{ComplexF64}` or `EmbeddedOperator`.
- **`_update_goal`** (`minimum_time_problem.jl:147`) accepts `AbstractPiccoloOperator` and passes it to the outer constructor. `GATES` entries are `Matrix{ComplexF64}`, so identity is preserved.
- **`SamplingTrajectory`** (`sampling_trajectory.jl:144`) constrains on `<:UnitaryTrajectory{P}` using only the first type parameter, unaffected by `G` removal.

## Verification

All 165 Piccolo tests pass, filtered to include every test item whose name contains "UnitaryTrajectory" plus all other existing test items that exercise `UnitaryTrajectory` indirectly (named trajectory conversion, rollout, integrators, problem templates).


jack-champagne:
I have dev installed this branch of piccolo with Piccolissimo and ran the testing suite. I am actually not entirely convinced that this needs to be marked breaking (as those who rely upon the previously weak type bound were abusing the type) and downstream doesn't seem to break

gennadiryan:
@jack-champagne Now that we're on v1.13.0 I'm good with merging this; I'm also not seeing any issues on the *issimo side of things (pertaining to this PR that is)

PR #178, opened 2026-04-30
feat: add calibration_targets kwarg to pulse problem templates

nguyenston:
## Summary

Adds `calibration_targets::Vector{Symbol} = Symbol[]` to all six pulse-problem entrypoints (`SplinePulseProblem`, `SmoothPulseProblem`, `BangBangPulseProblem` × single + MultiKet variants). Each name listed is pinned at its nominal value via `fix_global_variable!` so the QCP solve cannot drift it as a slack variable.

This implements the **intent layer** that was previously missing: a global declared in `calibration_targets` is a knob managed by an external calibration step (e.g. QILC), not a free NLP variable.

## Background

Until this PR, listing a global in `global_names` made it a free NLP variable in the QCP — even when the user intended it to be held at nominal during the control solve and managed externally. The optimizer would silently drift that global as a slack variable, polluting any downstream calibration baseline. Documented in [Intonato.jl/docs/onboarding.md §\"Globals: how they're defined and what their lifecycle should be\"](https://github.com/harmoniqs/Intonato.jl/blob/main/docs/onboarding.md).

The new kwarg makes the intent explicit:

```julia
qcp = SplinePulseProblem(qtraj, N; integrator=integ,
    global_names = [:χ_scale, :K_q_scale, :drive_scale],
    global_bounds = Dict(...),
    calibration_targets = [:χ_scale, :K_q_scale, :drive_scale],  # held at nominal
)
```

Default empty preserves current free-global behavior — no breaking change.

## Implementation

- New helper `apply_calibration_targets!(constraints, calibration_targets, traj; verbose=false)` in `_problem_templates.jl` mirrors the style of `add_global_bounds_constraints!`.
- Called after `add_global_bounds_constraints!` in each template so `calibration_targets` correctly supersedes `global_bounds` (the equality replaces any bounds on the same name).
- Errors clearly when a target name isn't in `traj.global_components`.

## Depends on

[harmoniqs/DirectTrajOpt.jl#82](https://github.com/harmoniqs/DirectTrajOpt.jl/pull/82) — adds the public `fix_global_variable!` this PR depends on. **Do not merge until DTO #82 is merged and a release is cut.**

## Test plan

- [x] New testitem `apply_calibration_targets! pins globals at nominal` (in `_problem_templates.jl`):
  - empty / no-op
  - pin at nominal
  - replace pre-existing GlobalBoundsConstraint
  - error on unknown global
  - multi-target
[...]

PR #185, opened 2026-05-01
feat: thread calibration_targets through SamplingProblem

nguyenston:
## Summary

Companion to #178 (`feat: add calibration_targets kwarg to pulse problem templates`). Adds `calibration_targets::Vector{Symbol} = Symbol[]` to `SamplingProblem`.

## Why this can't piggyback on the base qcp

The three base templates (Spline / Smooth / BangBang) added in #178 install `GlobalEqualityConstraint` entries on the QCP's constraint list. \`MinimumTimeProblem\` inherits those automatically — its body \`deepcopy\`s \`qcp.prob.constraints\` (line 102 of \`minimum_time_problem.jl\`).

\`SamplingProblem\` does **not**: it builds a fresh \`constraints = AbstractConstraint[]\` (line 205) on the new sampling trajectory rather than carrying the base qcp's constraints over (the trajectory shape changes — sampled states get duplicated per system — so the base constraints don't directly apply). Any calibration-target pins set on the base \`qcp\` are silently lost when wrapped in \`SamplingProblem\`. This PR makes the kwarg explicit so the caller restates them on the new sampled trajectory.

## Compat

Default empty preserves current behavior; existing callers see no change.

## Test plan

- [x] \`apply_calibration_targets!\` testitem still passes (9/9).
- [x] Smoke-test of SamplingProblem construction unaffected when kwarg defaults are used.

### DIFF
--- a/src/control/templates/_problem_templates.jl
+++ b/src/control/templates/_problem_templates.jl
@@ -227,6 +227,45 @@ function add_global_bounds_constraints!(
     end
 end
 
+"""
+    apply_calibration_targets!(constraints, calibration_targets, traj; verbose=false)
+
+For each global variable name in `calibration_targets`, pin it at its current
+value in `traj.global_data` via `fix_global_variable!`. Removes any existing
+`GlobalBoundsConstraint`/`GlobalEqualityConstraint` on the name to avoid MOI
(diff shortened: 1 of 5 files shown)


jack-champagne:
LGTM! @nguyenston

PR #189, opened 2026-05-04
fix: address Aqua and JET findings

jack-champagne:
## Summary

Triage and fix the 8 stale exports + 4 method ambiguities reported by Aqua and the 31 type-error findings reported by JET on the `ci/add-aqua-jet` baseline. Each finding category lands as its own commit so reviewers can read the chain top-to-bottom.

### Fixed

- **F1. Stale exports (Aqua, 8 → 1).** Removed `export open_rollout` and `export open_rollout_fidelity` from `src/quantum/dynamics.jl` (no definitions). Stopped re-exporting `plot_name!` from `QuantumObjectPlots` — `NamedTrajectories` owns the canonical export; the Piccolo-local definition stays in the submodule for `PiccoloMakieExt`'s qualified call.
- **F2. `Vector{Float64}(::Symbol)` in pulse constructors (JET, 6 → 0).** `ZeroOrderPulse` / `LinearSplinePulse` / `CubicSplinePulse` constructors gained an explicit `isa Symbol` arm that errors clearly on unknown symbols (only `:free` is supported alongside `nothing` and `Vector{<:Real}`).
- **F3. `CubicSplineInterpolation` in `dynamics.jl` (JET, 4 → 0).** That name is a deprecated `Interpolations.jl` symbol and is not reachable from this file. Replaced with `CubicHermiteSpline(traj, control_name)`, the documented `(traj, name)` 2-arg method provided by `NamedTrajectories.InterpolationsExt` over `DataInterpolations`.
- **F4. Other undefined names (JET, 4 → 0).** Reconstructed `SamplingTrajectory` directly inside `_update_goal(::SamplingTrajectory, ...)` instead of calling the undefined `update_base_trajectory`. Qualified the recursive `_update_system!` call as `Rollouts._update_system!`. Added an `else error(...)` arm for unknown `lab_frame_type` in `TransmonSystem`.
- **F5. SparseVector vs AbstractMatrix (JET, 3 → 0).** `OpenQuantumSystem`, `VariationalQuantumSystem`, `CompositeQuantumSystem` drift-less constructors built `spzeros(T, size(H_drives[1]))`. JET cannot prove the tuple has length 2, so it union-split into `SparseVector`. Destructure `(m, n) = size(...)` and call `spzeros(T, m, n)` — unambiguous matrix dispatch.
- **F6. Method ambiguities (Aqua, 4 → 0).** Two clusters: (a) `QuantumSystem(::_DriftInputs, ::Vector{<:AbstractDrive}, ...)` — added a 3-arg disambiguator that `invoke`s the pair-based path; (b) `UnitaryTrajectory(::QuantumSystem, ::G, ::G)` — narrowed the `(system, goal, T::Real)` method's `goal` to `Union{AbstractMatrix, AbstractPiccoloOperator}` (the only types ever passed in practice; verified via grep). Re-enabled `ambiguities` in `test/aqua.jl`.
- **F7. `default_algorithm(::OpenQuantumSystem)` (JET, 6 → 0).** Defined `default_algorithm(::OpenQuantumSystem) = Tsit5()`. Justification: the Lindblad generator is non-Hermitian by construction, and both `DensityTrajectory` and `MultiDensityTrajectory` constructors already default `algorithm = Tsit5()`. Added a smoke test that runs `rollout(::DensityTrajectory)` without an explicit algorithm and a unit test asserting the dispatch.
- **Misc JET inference cleanups.** Switched `if return_layers` to `if all_layers !== nothing` in `operator_algebra` (lets JET narrow the union). Bound `complex_control_norm_constraint_name` into a local before the `!== nothing` guard in `apply_piccolo_options!`. Tightened `add_global_bounds_constraints!` to `Union{Nothing, AbstractDict{Symbol, <:Any}}` and split `for pair in dict` with explicit `name::Symbol` binding. Replaced `collect(traj.global_names)` with `Symbol[traj.global_names...]` in `sampling_state_objective` (avoids `Vector{Union{}}` from empty tuples).

### Deferred / left as `broken=true`

- **F1 partial — `hessian_structure` (1 stale export).** This name is exported by three different `DirectTrajOpt` submodules (`CommonInterface`, `Constraints`, `Integrators`); the conflict makes it appear undefined at Piccolo's surface. Upstream issue, not Piccolo's to fix. `undefined_exports = (broken = true,)` is preserved with a comment.
- **F4 partial — `EnsembleSplineIntegrator` (1 JET finding).** Referenced inside `SplinePulseProblem(...)` when `integrator_type = :ensemble`, but no method is defined in Piccolo or any declared dep. The constructor itself errors users toward Piccolissimo's `SplineIntegrator`, suggesting this branch was meant to live behind an extension that hasn't been wired up. Left as the only JET `broken=true` item — see "Notes for human review" below.
- **F8.** Stdlib + test-extras compat policy — left as-is per instructions.

### Aqua before/after

| Category            | Before | After |
| ------------------- | ------ | ----- |
| `undefined_exports` | 8      | 1 (`hessian_structure`, upstream) |
| `ambiguities`       | 4      | 0 (re-enabled in `test/aqua.jl`)  |
| All other Aqua checks | pass | pass |

### JET before/after

| Category                                    | Before | After |
| ------------------------------------------- | ------ | ----- |
| `Vector{Float64}(::Symbol)` (pulse ctors)   | 6      | 0     |
| `CubicSplineInterpolation` undefined        | 4      | 0     |
| `default_algorithm(::OpenQuantumSystem)`    | 6      | 0     |
| `OpenQuantumSystem` / `VariationalQuantumSystem` / `CompositeQuantumSystem` SparseVector | 3 | 0 |
| Other undefined names (`update_base_trajectory`, `_update_system!`, `H_drift`, `EnsembleSplineIntegrator`) | 4 | 1 |
| `TransmonSystem` unbound `H_drift`          | 2      | 0     |
| `operator_algebra` `push!(::Nothing, ...)`  | 2      | 0     |
| `add_global_bounds_constraints!` / `apply_piccolo_options!` / `sampling_state_objective` inference | 3 | 0 |
| **Total**                                   | **31** | **1** |

### Test plan

- [x] `Aqua.test_all(Piccolo; deps_compat = (check_extras = false, ignore = [:Libdl]), undefined_exports = false)` — all checks pass except the upstream `hessian_structure`.
- [x] `JET.test_package(Piccolo; target_modules=(Piccolo,))` — 1 finding remaining (`EnsembleSplineIntegrator`), held as `broken=true`.
- [x] Smoke: `UnitaryTrajectory(sys, X_gate, T)`, `DensityTrajectory(open_sys, ρ0, ρg, T)`, `rollout(::DensityTrajectory)` without explicit algorithm — all green.
- [ ] Full `Pkg.test()` — running locally; may take 20–30 min. Will update once complete.

## Notes for human review (F7 + F4 partial)

- **F7 (`default_algorithm(::OpenQuantumSystem)`)**: I went the route of adding the method rather than deferring. Rationale: the open-system trajectory constructors already default `algorithm = Tsit5()` directly, so the new method just makes the indirect call site (`default_algorithm(qtraj.system)`) match. Also added a smoke test for `rollout(::DensityTrajectory)` so this stays exercised. If you'd prefer this be a separate per-system architectural decision, happy to revert.
- **F4 partial — `EnsembleSplineIntegrator`**: this looks like a Piccolissimo extension hook that was never finished. The error message in the surrounding code points users at `Piccolissimo: SplineIntegrator`. Two paths forward: (a) delete the `integrator_type == :ensemble` branch entirely, or (b) move it to a Piccolissimo extension. Either is a real call, so I left it alone. JET's `broken=true` will flip to a failure if someone later adds the symbol — that's the desired tripwire.

### DIFF
--- a/src/control/templates/sampling_problem.jl
+++ b/src/control/templates/sampling_problem.jl
@@ -227,7 +227,9 @@ end
 
 function _update_goal(qtraj::SamplingTrajectory, new_goal)
     new_base = _update_goal(qtraj.base_trajectory, new_goal)
-    return update_base_trajectory(qtraj, new_base)
+    # Reconstruct the SamplingTrajectory around the updated base — a thin
+    # SamplingTrajectory only owns (base_trajectory, systems, weights).
+    return SamplingTrajectory(new_base, qtraj.systems; weights = qtraj.weights)
 end
 
 function _final_fidelity_constraint(
--- a/src/control/templates/sampling_problem.jl
+++ b/src/control/templates/sampling_problem.jl
@@ -59,7 +59,11 @@ function sampling_state_objective(
     goal = get_goal(qtraj)
     # Use free-phase objective when globals exist and goal is EmbeddedOperator
     if traj.global_dim > 0 && goal isa EmbeddedOperator
-        θ_names = collect(traj.global_names)
+        # `Symbol[traj.global_names...]` rather than `collect(traj.global_names)` so
+        # the eltype is `Symbol` even when the source tuple is empty (JET infers
+        # `Vector{Union{}}` from `collect` on a possibly-empty tuple, which then
+        # fails to match `θ_names::AbstractVector{Symbol}` downstream).
+        θ_names = Symbol[traj.global_names...]
         U_goal_fn = _make_free_phase_goal(goal)
         return UnitaryFreePhaseInfidelityObjective(
             U_goal_fn,
--- a/src/quantum/dynamics.jl
+++ b/src/quantum/dynamics.jl
@@ -517,7 +517,7 @@ function rollout_fidelity(
     elseif interpolation == :linear
         u = LinearInterpolation(traj, control_name)
     elseif interpolation == :cubic
-        u = CubicSplineInterpolation(traj, control_name)
+        u = CubicHermiteSpline(traj, control_name)
     else
         error(
             "Unknown interpolation method: $(interpolation). Use :constant, :linear, or :cubic",
@@ -572,7 +572,7 @@ function unitary_rollout_fidelity(
     elseif interpolation == :linear
         u = LinearInterpolation(traj, control_name)
     elseif interpolation == :cubic
-        u = CubicSplineInterpolation(traj, control_name)
+        u = CubicHermiteSpline(traj, control_name)
     else
         error(
             "Unknown interpolation method: $(interpolation). Use :constant, :linear, or :cubic",
@@ -609,7 +609,7 @@ function unitary_rollout(
     elseif interpolation == :linear
         u = LinearInterpolation(traj, control_name)
     elseif interpolation == :cubic
-        u = CubicSplineInterpolation(traj, control_name)
+        u = CubicHermiteSpline(traj, control_name)
     else
         error(
             "Unknown interpolation method: $(interpolation). Use :constant, :linear, or :cubic",
@@ -670,7 +670,7 @@ function ket_rollout(
     elseif interpolation == :linear
         u = LinearInterpolation(traj, control_name)
     elseif interpolation == :cubic
-        u = CubicSplineInterpolation(traj, control_name)
+        u = CubicHermiteSpline(traj, control_name)
     else
         error(
             "Unknown interpolation method: $(interpolation). Use :constant, :linear, or :cubic",
--- a/src/quantum/dynamics.jl
+++ b/src/quantum/dynamics.jl
@@ -67,8 +67,6 @@ export ket_rollout
 export ket_rollout_fidelity
 export unitary_rollout
 export unitary_rollout_fidelity
-export open_rollout
-export open_rollout_fidelity
 export update_global_params!
 
 export KetODEProblem
--- a/src/quantum/systems/quantum_systems.jl
+++ b/src/quantum/systems/quantum_systems.jl
@@ -593,6 +593,35 @@ function QuantumSystem(
     )
 end
 
+# Disambiguator: a `(drift::_DriftInputs, drives::Vector{<:AbstractDrive}, ...)` call
+# matches both the pair-based method above and the structured-drift method below.
+# Both methods share the same intent for this input shape; route through the
+# pair-based path (it handles `Vector{<:AbstractDrive}` natively in its drive loop)
+# so dispatch is unambiguous.
+function QuantumSystem(
+    drift::_DriftInputs,
+    drives::Vector{<:AbstractDrive},
+    drive_bounds::Vector{<:Union{Tuple{Float64,Float64},Float64}};
+    time_dependent::Bool = false,
+    global_params::NamedTuple = NamedTuple(),
+    hermitian::Bool = true,
+)
+    return invoke(
+        QuantumSystem,
+        Tuple{
+            _DriftInputs,
+            AbstractVector,
+            Vector{<:Union{Tuple{Float64,Float64},Float64}},
+        },
+        drift,
+        drives,
+        drive_bounds;
+        time_dependent = time_dependent,
+        global_params = global_params,
+        hermitian = hermitian,
+    )
+end
+
 """
     QuantumSystem(
         H_drift,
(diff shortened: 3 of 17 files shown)


src/control/templates/sampling_problem.jl
```diff
@@ -227,7 +231,9 @@ end
 
 function _update_goal(qtraj::SamplingTrajectory, new_goal)
     new_base = _update_goal(qtraj.base_trajectory, new_goal)
-    return update_base_trajectory(qtraj, new_base)
+    # Reconstruct the SamplingTrajectory around the updated base — a thin
+    # SamplingTrajectory only owns (base_trajectory, systems, weights).
+    return SamplingTrajectory(new_base, qtraj.systems; weights = qtraj.weights)
```
jack-champagne, line 236: Need a closer look to determine how we want to do this

src/quantum/systems/quantum_systems.jl
```diff
@@ -593,6 +593,31 @@ function QuantumSystem(
...
+        QuantumSystem,
+        Tuple{_DriftInputs,AbstractVector,Vector{<:Union{Tuple{Float64,Float64},Float64}}},
+        drift,
+        drives,
+        drive_bounds;
+        time_dependent = time_dependent,
+        global_params = global_params,
+        hermitian = hermitian,
+    )
+end
```
jack-champagne, line 619: This needs a closer look too, this shim for the constructor alludes to a possible type issue on the constructor that we should fix with other stuff...

src/quantum/trajectories/_quantum_trajectories.jl
```diff
@@ -39,15 +39,49 @@ import ..QuantumSystems: default_algorithm
...
+matches the explicit `algorithm = Tsit5()` default in the
+`DensityTrajectory` / `MultiDensityTrajectory` constructors.
 """
 function default_algorithm(sys::QuantumSystem)
     return sys.hermitian ? MagnusAdapt4() : Tsit5()
 end
 
+function default_algorithm(::OpenQuantumSystem)
+    return Tsit5()
+end
```
jack-champagne, line 69: This is an important change that I am *fairly* certain we should integrate unless you have a different integrator in mind. @aarontrowbridge @gennadiryan @andgoldschmidt

src/quantum/trajectories/unitary_trajectory.jl
```diff
@@ -90,13 +90,13 @@ Convenience constructor that creates a zero pulse of duration T.
 """
 function UnitaryTrajectory(
     system::QuantumSystem,
-    goal::G,
+    goal::Union{AbstractMatrix,AbstractPiccoloOperator},
```
jack-champagne: Was this type scoping doing anything before? Where was G being defined? Was it strict enough?

src/quantum/dynamics.jl
```diff
@@ -519,7 +517,7 @@ function rollout_fidelity(
     elseif interpolation == :linear
         u = LinearInterpolation(traj, control_name)
     elseif interpolation == :cubic
-        u = CubicSplineInterpolation(traj, control_name)
+        u = CubicHermiteSpline(traj, control_name)
```
jack-champagne, line 520: Not sure if this is actually the right type, and if it isn't then we should have flagged this in the tests. Will need to inspect to validate

src/visualizations/quantum_objects/_quantum_object_plots.jl
```diff
@@ -3,7 +3,9 @@ module QuantumObjectPlots
 export plot_unitary_populations
 export plot_state_populations
 export plot_weyl_trajectory
-export plot_name!
+# plot_name! is defined here for use within QuantumObjectPlots / PiccoloMakieExt; the
```
jack-champagne, line 6: need to check this too - have had issues in the past with fiddling with these exports with the extension package for (Cairo)Makie on NamedTrajectories.jl

jack-champagne (commented):
Needs a bit of work, leaving this in draft until we can get this all addressed.

PR #194, opened 2026-05-08
Validate nonlinear drives with appended global parameters

mabuni1998:
## Summary

Fix nonlinear-drive validation for systems that append global parameters to the optimization vector.

`NonlinearDrive` coefficients can depend on the full `u = [controls..., globals...]` vector, but the validation logic in `Piccolo` was only checking the first `n_drives` entries. That causes validation to fail for valid drives that reference appended global parameters.

## Problem

When constructing a `QuantumSystem` or `OpenQuantumSystem` with `global_params`, the effective input to a drive coefficient is longer than the physical control dimension.

Before this change:
- `validate_drive_jacobian` and `validate_drive_hessian` sampled vectors of length `n_drives`
- constructor-side validation passed only `n_drives`
- any `NonlinearDrive` that referenced appended globals could throw bounds errors during validation

This showed up for drives with coefficients like:
- `u[i] * u[j] * u[global_idx]`

where `global_idx > n_drives`.

## Changes

- Rename the validator argument from `n_controls` to `u_dim` to reflect its real meaning
- Update `validate_drive_jacobian` to sample and validate over the full `u_dim`
- Update `validate_drive_hessian` the same way
- Pass `n_drives + length(global_params)` into nonlinear-drive validation from:
  - `QuantumSystem`
  - the typed-drive `QuantumSystem` constructor path
  - `OpenQuantumSystem`

## Notes

This change only affects validation behavior. It does not change the runtime definition of the Hamiltonian or drive coefficients themselves.

### DIFF
--- a/src/quantum/systems/drives.jl
+++ b/src/quantum/systems/drives.jl
@@ -698,6 +698,46 @@ end
     @test_throws AssertionError validate_drive_hessian(d_wrong, 2)
 end
 
+@testitem "validate_drive: u_dim extending past n_drives (global params)" begin
+    # Regression: when global_params are appended, callers pass
+    # u_dim = n_drives + length(global_params) so coefficients that read into
+    # the global-parameter slots can be validated without BoundsError.
+    using Piccolo
+    using SparseArrays
+
+    H = sparse([0.0+0im 1.0+0im; 1.0+0im 0.0+0im])
+
+    # Coefficient reads u[3] — i.e. the appended global slot when n_drives = 2
+    # and one global parameter is present.
+    d_auto = NonlinearDrive(H, u -> u[1] * u[2] * u[3])
+    validate_drive_jacobian(d_auto, 3)  # must not BoundsError
+    validate_drive_hessian(d_auto, 3)
+
+    # Explicit jacobian + hessian agreeing with ForwardDiff across the full
+    # u_dim = 3 must pass.
+    d_correct = NonlinearDrive(
+        H,
+        u -> u[1] * u[2] * u[3],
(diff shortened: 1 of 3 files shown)


jack-champagne:
+1 to this approach over the parallel fix in #196. Threading `u_dim = n_drives + length(global_params)` through from the call sites is the more honest semantic — the validator's job is to spot-check the coefficient over the full `u` vector that the integrator will actually feed it, and the call site is the only place that knows that length. #196's `max(n_controls, maximum(active_controls))` trick inside the validator only fixes the case where the user explicitly declares the global slots in `active_controls`; it leaves the documented "empty = all controls" form (the recommended 2-arg `NonlinearDrive(H, coeff)`) still BoundsErroring when globals are appended.

Pushed two regression testitems on top of this branch (`1202eb7`, co-authored with Raghav from #196):

- `validate_drive: u_dim extending past n_drives (global params)` in `drives.jl` — direct validator test at `u_dim = 3` with a coefficient that reads `u[3]`, covering the auto-derivative path, a correct explicit jacobian+hessian, and a wrong jacobian at `j = 3` (which must still assert).
- `QuantumSystem: NonlinearDrive reading appended global params` in `quantum_systems.jl` — constructor-level test that locks in the new wiring (`n_drives + length(global_params)` passed to `validate_drive_jacobian`). Reverting the call-site change to `n_drives` would break this test.

Both pass locally. Happy to also port the `OpenQuantumSystem` call site into a matching test if you'd like coverage symmetry, but the wiring is identical so the existing pass should be sufficient.

PR #195, opened 2026-05-09
Add pulse evolution animations

### LINKED ISSUE #59: [unitaryDESIGN] Animation of a Pulse Evolving with Parameters
## Description

Create an **animation of a control pulse** that evolves as parameters change. Piccolo.jl already has basic animation infrastructure (`animate_figure`, `animate_name` in `PiccoloMakieExt.jl`), but there's an opportunity to create richer, more interactive pulse animations.

**Bounty: $100**

## Why it matters

Pulse visualization is central to quantum optimal control — researchers need to see how pulses change during optimization or across parameter sweeps. An engaging animation could be used on the [Harmoniqs website](https://harmoniqs.co) and would improve the developer experience.

## Relevant code & references

- [`animate_figure`](https://github.com/harmoniqs/Piccolo.jl/blob/main/ext/PiccoloMakieExt.jl) — Existing animation helper (supports `:inline` and `:record` modes)
- [`animate_name`](https://github.com/harmoniqs/Piccolo.jl/blob/main/ext/PiccoloMakieExt.jl) — Animates trajectory variable evolution
- [Visualization guide](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/guides/visualization.jl) — Shows control plotting, phase space, fidelity evolution
- [GLMakie](https://docs.makie.org/stable/explanations/backends/glmakie/) — Interactive Makie backend (recommended for real-time)

## Scope

**In scope:**
- Animate a pulse evolving over time (e.g., showing the pulse "drawing" itself as time progresses)
- Show corresponding quantum state or population evolution alongside the pulse
- Bonus: sliders/interactivity to change pulse parameters (amplitude, frequency, duration) and see the effect in real-time using `GLMakie` Observables

**Out of scope:**
- Full GUI application
- Modifications to the core optimization solver

## Potential outcomes

- Animated `.mp4` or `.gif` of pulse evolution suitable for website/blog
- Interactive GLMakie dashboard with parameter sliders (stretch goal)
- Pulses changing in real-time

grantf04:
Closes #59.

## Summary
- adds `animate_pulse` as a Makie-backed visualization API for all `AbstractPulse` types
- supports optional population/state traces below the pulse animation
- documents pulse-only and pulse-plus-population examples in the visualization guide

## Validation
- `git diff --check`

Note: I could not run the Julia test suite locally because Julia is not installed in this environment. I will follow up on CI results.

### DIFF
--- a/docs/literate/guides/visualization.jl
+++ b/docs/literate/guides/visualization.jl
@@ -148,6 +148,47 @@ fig = plot_pulse_IQ(pulse_iq; title = "IQ view (Ω, α)")
 
 fig = plot_pulse_phases(pulse_iq; title = "Polar view (|·|, ∠·)")
 
+# ## Pulse Animations
+#
+# `animate_pulse` shows a pulse drawing itself over time. Use `mode = :record`
+# with a Makie backend that can record frames when you want an `.mp4` or `.gif`
+# for a website or notebook; use `mode = :inline` with GLMakie for a looping
+# interactive preview.
+
+animation_pulse = GaussianPulse([0.8, 0.45], T / 6, T)
+
+fig = animate_pulse(
+    animation_pulse;
+    mode = :inline, # use mode = :record and filename = "pulse_evolution.mp4" to export
+    fps = 24,
+    n_samples = 120,
+    labels = ["Ω_x", "Ω_y"],
+    title = "Gaussian control pulse",
+)
+
+# You can add state or population traces underneath the controls by passing a
+# matrix whose rows are populations and whose columns are time samples. This is
+# useful for showing the control pulse and the quantum response in the same
+# animation.
+
+population_times = collect(range(0, T, length = 120))
+population_trace = vcat(
+    reshape(cos.(π .* population_times ./ (2T)) .^ 2, 1, :),
+    reshape(sin.(π .* population_times ./ (2T)) .^ 2, 1, :),
+)
+
+fig = animate_pulse(
+    animation_pulse;
+    mode = :inline, # use mode = :record and filename = "pulse_with_populations.mp4" to export
+    fps = 24,
+    n_samples = 120,
+    labels = ["Ω_x", "Ω_y"],
+    populations = population_trace,
+    population_times,
+    population_labels = ["|0⟩", "|1⟩"],
+    title = "Pulse with population evolution",
+)
+
 # ## Custom Plotting
 #
 # For full control, extract trajectory data and use Makie directly.
--- a/ext/PiccoloMakieExt.jl
+++ b/ext/PiccoloMakieExt.jl
@@ -106,6 +106,252 @@ function Piccolo.animate_name(
 end
 
 
+"""
+    animate_pulse(
+        pulse::AbstractPulse;
+        fps::Int = 24,
+        mode::Symbol = :inline,
+        filename::String = "pulse_animation.mp4",
+        n_samples::Int = 240,
+        labels = nothing,
+        populations = nothing,
+        population_times = nothing,
+        population_labels = nothing,
+        title::AbstractString = "Pulse evolution"
+    )
+
+Animate a pulse being drawn over time, with optional population/state traces in
+a second panel. The pulse is sampled uniformly so this works with every
+`AbstractPulse` implementation.
+"""
+function Piccolo.animate_pulse(
+    pulse::Piccolo.AbstractPulse;
+    fps::Int = 24,
+    mode::Symbol = :inline,
+    filename::String = "pulse_animation.mp4",
+    n_samples::Int = 240,
+    labels = nothing,
+    populations = nothing,
+    population_times = nothing,
+    population_labels = nothing,
+    title::AbstractString = "Pulse evolution",
+)
+    n_samples >= 2 || throw(ArgumentError("n_samples must be at least 2."))
+
+    pulse_times = collect(range(0.0, Piccolo.duration(pulse), length = n_samples))
+    pulse_values = Piccolo.sample(pulse, pulse_times)
+    n_drive = Piccolo.n_drives(pulse)
+
+    size(pulse_values, 1) == n_drive ||
+        throw(ArgumentError("sample(pulse, times) must return one row per drive."))
+    size(pulse_values, 2) == n_samples ||
+        throw(ArgumentError("sample(pulse, times) must return one column per sample."))
+
+    drive_labels = if isnothing(labels)
+        _pulse_drive_labels(pulse, n_drive)
+    else
+        length(labels) == n_drive ||
+            throw(ArgumentError("labels must contain one entry per pulse drive."))
+        collect(labels)
+    end
+
+    has_populations = !isnothing(populations)
+    population_values = nothing
+    pop_times = nothing
+    pop_labels = nothing
+    n_pop = 0
+
+    if has_populations
+        population_values = Matrix(populations)
+        size(population_values, 2) >= 2 ||
+            throw(ArgumentError("populations must have at least two time samples."))
+
+        n_pop = size(population_values, 1)
+        pop_times =
+            isnothing(population_times) ?
+            collect(
+                range(
+                    pulse_times[1],
+                    pulse_times[end],
+                    length = size(population_values, 2),
+                ),
+            ) : collect(population_times)
+
+        length(pop_times) == size(population_values, 2) || throw(
+            ArgumentError("population_times must match the number of population samples."),
+        )
+
+        pop_labels = if isnothing(population_labels)
+            ["Population $i" for i = 1:n_pop]
+        else
+            length(population_labels) == n_pop || throw(
+                ArgumentError(
+                    "population_labels must contain one entry per population row.",
+                ),
+            )
+            collect(population_labels)
+        end
+    end
+
+    fig = Figure(size = has_populations ? (900, 620) : (900, 380))
+    pulse_axis = Axis(
+        fig[1, 1],
+        xlabel = has_populations ? "" : "Time",
+        xticklabelsvisible = !has_populations,
+        ylabel = "Amplitude",
+        title = title,
+        titlealign = :left,
+    )
+
+    pulse_y_min, pulse_y_max = _padded_extrema(pulse_values)
+    xlims!(pulse_axis, pulse_times[1], pulse_times[end])
+    ylims!(pulse_axis, pulse_y_min, pulse_y_max)
+    hlines!(pulse_axis, [0.0], color = (:gray, 0.45), linestyle = :dash, linewidth = 1)
+
+    colors = Makie.wong_colors()
+    pulse_xs = [Observable(pulse_times[1:1]) for _ = 1:n_drive]
+    pulse_ys = [Observable(vec(pulse_values[i, 1:1])) for i = 1:n_drive]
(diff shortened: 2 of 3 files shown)


grantf04:
Follow-up validation after the latest push:

- `git diff --check`
- temp Julia 1.11.7 env with `CairoMakie`: `animate_pulse(... mode=:inline, populations=...)`
- temp Julia 1.11.7 env with `CairoMakie`: `animate_pulse(... mode=:record, filename=tempname()*".mp4")`

I also fixed the default labels for analytic pulses that do not expose `drive_name`.

ext/PiccoloMakieExt.jl
```diff
@@ -106,6 +106,252 @@ function Piccolo.animate_name(
...
+
+    colors = Makie.wong_colors()
+    pulse_xs = [Observable(pulse_times[1:1]) for _ = 1:n_drive]
+    pulse_ys = [Observable(vec(pulse_values[i, 1:1])) for i = 1:n_drive]
+    marker_xs = [Observable([pulse_times[1]]) for _ = 1:n_drive]
+    marker_ys = [Observable([pulse_values[i, 1]]) for i = 1:n_drive]
+
+    for i = 1:n_drive
+        color = colors[mod1(i, length(colors))]
+        lines!(
```
jack-champagne, line 219: **Correctness:** rendering as dense `lines!` regardless of pulse type means `ZeroOrderPulse` will animate as a smooth interpolated curve instead of stairs, and `LinearSplinePulse` will show dense samples through knots instead of knot-to-knot segments. The static `plot_pulse!` already gets this right via type dispatch (`src/visualizations/quantum_objects/pulse_plots.jl:323` for ZOH, `:366` for linear, `:408` for cubic, `:480` for analytic). The architecture suggestion in the top-level comment fixes this by delegating; if you don't take that route, at minimum this needs per-type branching here.

ext/PiccoloMakieExt.jl
```diff
@@ -106,6 +106,252 @@ function Piccolo.animate_name(
...
+        title = title,
+        titlealign = :left,
+    )
+
+    pulse_y_min, pulse_y_max = _padded_extrema(pulse_values)
+    xlims!(pulse_axis, pulse_times[1], pulse_times[end])
+    ylims!(pulse_axis, pulse_y_min, pulse_y_max)
+    hlines!(pulse_axis, [0.0], color = (:gray, 0.45), linestyle = :dash, linewidth = 1)
+
+    colors = Makie.wong_colors()
```
jack-champagne, line 211: **Duplication with existing helpers** in `src/visualizations/quantum_objects/pulse_plots.jl`:

- `Makie.wong_colors()` + `colors[mod1(i, length(colors))]` → use `_drive_color(i)` (`pulse_plots.jl:61`), which is theme-aware and falls back to Wong.
- `_pulse_drive_labels` (defined further down) → `_default_drive_labels(pulse)` (`pulse_plots.jl:804`) returns `LaTeXString`s with proper math subscripts via `latexstring(base, "_{", i, "}")`.
- `hlines!(pulse_axis, [0.0], color = (:gray, 0.45), ...)` → `_draw_pulse_subplot!` (`pulse_plots.jl:516`) already adds this zero-reference using `_theme_neutral()` so it stays visible on dark themes.
- No bounds support in `animate_pulse`. `_draw_bounds!` (`pulse_plots.jl:577`) handles this for the static plotter.

If `plot_pulse!` does the rendering (per top-level comment), all of this comes along for free.

ext/PiccoloMakieExt.jl
```diff
@@ -106,6 +106,252 @@ function Piccolo.animate_name(
...
+    population_labels = nothing,
+    title::AbstractString = "Pulse evolution",
+)
+    n_samples >= 2 || throw(ArgumentError("n_samples must be at least 2."))
+
+    pulse_times = collect(range(0.0, Piccolo.duration(pulse), length = n_samples))
+    pulse_values = Piccolo.sample(pulse, pulse_times)
+    n_drive = Piccolo.n_drives(pulse)
+
+    size(pulse_values, 1) == n_drive ||
```
jack-champagne, line 145: **Defensive code that shouldn't be needed:**

1. The `size(pulse_values, 1) == n_drive` / `size(pulse_values, 2) == n_samples` asserts here are validating the `sample` interface contract from outside. If a subtype violates that contract, it's a bug in the pulse type — those assertions belong in a test suite for `AbstractPulse` implementations, not at the animation call site.
2. The `try/catch` around `Piccolo.drive_name(pulse)` in `_pulse_drive_labels` below swallows real errors silently — `drive_name` is part of the `AbstractPulse` interface (`src/quantum/primitives/pulses.jl:80`) and should always work. The fallback also relies on a string equality (`base == "Drive"`) to pick between two label formats, which silently changes behavior if a pulse happens to be named `:Drive`.

Both can be deleted once `plot_pulse!` is doing the rendering.

ext/PiccoloMakieExt.jl
```diff
@@ -122,4 +368,25 @@ end
 end
 
 
+@testitem "Test animate_pulse for AbstractPulse" begin
```
jack-champagne, line 371: Could you add a few more tests while you're in there? The current test exercises Gaussian + populations + `:inline`; it'd be great to also cover:

- A `:record`-mode test writing to `tempname() * ".mp4"` so the record path has CI coverage.
- A no-populations test — the populations-off branch currently isn't exercised in CI.
- At least one non-Gaussian pulse, e.g. `ZeroOrderPulse(controls, times)` or `LinearSplinePulse(...)`. This would also serve as a regression test for the type-aware rendering once the architecture change in the top-level comment lands.
- An `n_samples = 1` case (currently throws) and a label-length mismatch case, just to lock the error contract.

jack-champagne (commented):
Thanks for taking this on — closes the in-scope items from #59, and observable-based updates is the right pattern.

Before merge I'd like to propose composing this with the existing `plot_pulse!` machinery instead of reimplementing rendering. The cleanest shape is to make `plot_pulse!` itself animation-aware:

```julia
plot_pulse!(ax, pulse; ..., t_max::Union{Nothing,Observable{Float64}} = nothing)
```

When `nothing` (default), behavior is unchanged. When set, thread `t_max` into each `_plot_pulse_type!` method. The existing per-type dispatch already chooses the right rendering primitive (stairs / knot-lines / dense curve) — we just want it to produce observable-driven primitives that respect the visible window.

The nice simplification: **`(pulse::AbstractPulse)(t::Real) = evaluate(pulse, t)` at `src/quantum/primitives/pulses.jl:60` already does the type-specific math**. Held-value for `ZeroOrderPulse`, linear-interp for `LinearSplinePulse`, cubic for `CubicSplinePulse`, analytic for Gaussian/Erf. So a single reveal helper works for every type — the synthetic endpoint at `t_max` is just `pulse(t_max[])`:

```julia
# Shared helper — works for any pulse type because pulse(t) handles the math
function _reveal_xy(pulse, render_times, drive_index, t_max)
    k = searchsortedlast(render_times, t_max)
    xs = k == 0 ? [t_max] : vcat(render_times[1:k], t_max)
    ys = [pulse(t)[drive_index] for t in xs]
    return xs, ys
end
```

Each `_plot_pulse_type!` keeps its existing logic for choosing `render_times` (knot times for ZOH/linear, dense for cubic/analytic) and just routes xs/ys through this helper when `t_max` is non-nothing:

```julia
function _plot_pulse_type!(ax, pulse::ZeroOrderPulse; ..., t_max, ...)
    knot_times = collect(get_knot_times(pulse))
    knot_vals  = sample(pulse, knot_times)
    for (ci, i) in enumerate(drive_indices)
        xs, ys = if isnothing(t_max)
            knot_times, knot_vals[i, :]
        else
            (@lift _reveal_xy(pulse, knot_times, i, $t_max)[1]),
            (@lift _reveal_xy(pulse, knot_times, i, $t_max)[2])
        end
        stairs!(ax, xs, ys; color = colors[ci], step = :post, linewidth = 2)
        ...
    end
end
```

Then `animate_pulse` collapses to pure orchestration:

```julia
t_max = Observable(0.0)
# Type-aware ghost (low alpha) — full dispatch, no observable
plot_pulse!(ax, pulse; show_knots=false,
            colors=[(_drive_color(i), 0.18) for i in 1:n_drives(pulse)])
# Type-aware reveal — same dispatch, observable-driven
plot_pulse!(ax, pulse; t_max)
# Leading-edge marker — pulse(t_max) gives the right value automatically
for i in 1:n_drives(pulse)
    pt = @lift Point2f($t_max, pulse($t_max)[i])
    scatter!(ax, @lift([$pt]); color = _drive_color(i), markersize = 10)
end
animate_figure(fig, 1:n_samples, i -> (t_max[] = pulse_times[i]); mode, fps, filename)
```

Net effect:
- `ZeroOrderPulse` animates as actual stairs, `LinearSplinePulse` knot-to-knot, cubic/analytic stay smooth — all via the existing dispatch.
- Color, label, theme, zero-line, and bound logic stays in one place.
- `animate_pulse` no longer touches `lines!`, `stairs!`, or per-type detection.
- The `plot_pulse!` kwarg defaults to `nothing` — fully backward-compatible.

Smaller concerns split out as separate comments below.

docs/literate/guides/visualization.jl
```diff
@@ -148,6 +148,47 @@ fig = plot_pulse_IQ(pulse_iq; title = "IQ view (Ω, α)")
...
+# `animate_pulse` shows a pulse drawing itself over time. Use `mode = :record`
+# with a Makie backend that can record frames when you want an `.mp4` or `.gif`
+# for a website or notebook; use `mode = :inline` with GLMakie for a looping
+# interactive preview.
+
+animation_pulse = GaussianPulse([0.8, 0.45], T / 6, T)
+
+fig = animate_pulse(
+    animation_pulse;
+    mode = :inline, # use mode = :record and filename = "pulse_evolution.mp4" to export
```
jack-champagne, line 162: Could you produce the animation and drop a short clip (mp4 or gif) in a reply to this comment so we can see the result? Use `mode = :record, filename = "pulse_evolution.mp4"` with one of the doc examples — the `:inline` path won't actually animate under CairoMakie (see the warning at `animate_figure` line 32). If your hardware can't run GLMakie or the record path, no worries — let me know and I can produce the clip myself.

Don't worry about embedding the video into the docs as part of this PR — we'll handle docs integration separately.

jack-champagne:
Hey @grantf04, this looks a lot of the way there. Do you have any questions on my comments? (Don't worry about the codecov flag, the other CI jobs passed which is what I am looking for).

jack-champagne:
Hi @grantf04 — thank you for this. The core ideas here were right: the `animate_pulse` API surface, the optional population panel, the observable-driven update pattern, and the docs example are all good, and they live on in the replacement PR.

I haven't heard back on the review comments for about a month and `main` has since moved (this branch is now conflicting), so I'm closing this in favor of #240, which picks up your work and finishes the architecture direction from the review — delegating rendering to the existing `plot_pulse!` machinery so each pulse type animates with its native primitive (stairs for `ZeroOrderPulse`, knot-to-knot for linear, smooth for cubic/analytic), with the duplicated color/label/zero-line handling and the defensive code removed.

Your contribution is carried over and credited (`Co-authored-by`) on the commit in #240. Closing this one purely for staleness — not a reflection on the work. Thanks again for taking #59 on, and please feel free to review #240.

PR #196, opened 2026-05-11
fix(drives): extend NonlinearDrive validator u-vector for active_controls beyond n_controls

Rchari1:
## Summary

`validate_drive_jacobian` and `validate_drive_hessian` generated a test vector of length `n_controls` and ignored `d.active_controls`. Drives whose `active_controls` reach beyond `n_controls` — e.g. a `NonlinearDrive` whose coefficient depends on appended global parameters — would throw `BoundsError` during ForwardDiff evaluation, blocking `QuantumSystem` construction.

This fixes the validator to extend its test vector to `max(n_controls, maximum(d.active_controls))` and iterate over the full extended range, so inactive indices are also checked (they must return zero gradient/Hessian, which is itself a useful invariant).

## Reproducer

```julia
using Piccolo
H = sparse([0.0+0im 1.0+0im; 1.0+0im 0.0+0im])
# A drive that reads u[3] even though only 2 physical controls are declared:
d = NonlinearDrive(H, u -> u[1] * u[2] * u[3]; active_controls = [1, 2, 3])
QuantumSystem(H, [d], [(0.0, 1.0), (0.0, 1.0)];
    global_params = (g1 = 0.5,))  # BoundsError on `main`, OK on this branch
```

## What changed

`src/quantum/systems/drives.jl`: introduce `_validation_u_dim(d, n_controls)` helper; use it in both `validate_drive_jacobian` and `validate_drive_hessian`. Behavior is bit-for-bit unchanged when `active_controls` is empty or stays within `1:n_controls`.

## Test plan

- [x] Repro above passes against this branch
- [x] Built a 6-cavity Kerr + cross-coupling `QuantumSystem` (21 `NonlinearDrive` and `LinearDrive` terms, 15 globals); previously bounds-errored, now constructs cleanly
- [x] Running a full Altissimo curriculum (Stage 1 CPU + Stage 2 GPU at N=30, K=8) against this branch via Pkg.develop; Stage 1 ran to completion across 6 outer × ~30 inner Newton-CG iterations (≫ thousands of derivative evaluations through the patched validator path) without validator errors. Stage 2 in progress at time of writing.
- [ ] Existing Piccolo test suite — left for CI to confirm

## Notes

- Marked draft pending CI green and any review feedback on the exact validator semantics (e.g. whether inactive-index zero-gradient should also be loop-tested — current patch does both; happy to scope down if preferred).

### DIFF
--- a/src/quantum/systems/drives.jl
+++ b/src/quantum/systems/drives.jl
@@ -713,6 +713,47 @@ end
     @test_throws AssertionError validate_drive_hessian(d_wrong, 2)
 end
 
+@testitem "validate_drive: active_controls reaching beyond n_controls" begin
+    # Regression for the BoundsError that occurred when a NonlinearDrive's
+    # coefficient reads indices beyond `n_controls` (e.g. appended global
+    # parameters) and declares them via `active_controls`.
+    using Piccolo
+    using SparseArrays
+
+    H = sparse([0.0+0im 1.0+0im; 1.0+0im 0.0+0im])
+
+    # Auto-jacobian/hessian: coeff reads u[3] while n_controls = 2.
+    d_auto = NonlinearDrive(H, u -> u[1] * u[2] * u[3]; active_controls = [1, 2, 3])
+    validate_drive_jacobian(d_auto, 2)  # must not BoundsError
+    validate_drive_hessian(d_auto, 2)
+
+    # Explicit jacobian + hessian that agree with ForwardDiff over the
+    # extended range pass.
+    d_correct = NonlinearDrive(
+        H,
+        u -> u[1] * u[2] * u[3],
+        (u, j) ->
+            j == 1 ? u[2] * u[3] : j == 2 ? u[1] * u[3] : j == 3 ? u[1] * u[2] : 0.0;
+        active_controls = [1, 2, 3],
+        coeff_hess = (u, i, j) ->
+            (i == 1 && j == 2) || (i == 2 && j == 1) ? u[3] :
+            (i == 1 && j == 3) || (i == 3 && j == 1) ? u[2] :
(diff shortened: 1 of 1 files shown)


jack-champagne:
Approved — clean, well-scoped fix. Pushed a regression testitem (`2ac7741`) modeled on the PR-description reproducer: a `NonlinearDrive(u -> u[1] * u[2] * u[3]; active_controls = [1, 2, 3])` at `n_controls = 2`, covering the auto-derivative path, a correct explicit jacobian+hessian, and a wrong explicit jacobian at `j = 3` (which the extended-range loop must catch). Ran locally — passes on this branch.

jack-champagne (approved):

jack-champagne:
Walking back the unqualified approval above — #194 (mabuni1998, opened 2026-05-08) is solving the same bug with a different and more complete approach: rename the validator's `n_controls` → `u_dim` and have call sites pass `n_drives + length(global_params)`. That fixes the case this PR leaves broken: a `NonlinearDrive(H, u -> u[3])` with **empty `active_controls`** (the documented "all controls, no sparsity info" form) and an appended global parameter still BoundsErrors under #196, since `_validation_u_dim` short-circuits to `n_controls`.

That's not a hypothetical — the 2-arg `NonlinearDrive(H, coeff)` constructor takes no `active_controls` and is the recommended form in the docstring, so this case is easy to hit by accident.

Given that, my suggestion is to close this PR in favor of #194 and let that one land. I've ported the regression test from this branch onto #194's branch (with you credited as `Co-Authored-By` on the commit) so the coverage doesn't get lost — see https://github.com/harmoniqs/Piccolo.jl/pull/194.

Sorry for the back-and-forth — I should have searched for an existing PR before approving this one.

jack-champagne:
Merged #194 , closing

PR #205, opened 2026-05-15
ci: shard testitem runs across 4 parallel jobs

jack-champagne:
## Summary

Cuts wallclock CI time by sharding `@run_package_tests` across 4 parallel GitHub matrix jobs. Each job runs ~1/4 of the discovered testitems, selected by a stable hash of `(filename, name)`.

- `test/runtests.jl`: respects `TEST_SHARD` / `TOTAL_SHARDS` env vars. When unset (local `Pkg.test`, REPL, VS Code), behavior is unchanged — runs every testitem.
- `.github/workflows/CI.yml`: adds `shard: ['1','2','3','4']` to the matrix, sets `TOTAL_SHARDS=4`.

Expected impact: ~3x wallclock speedup once Julia precompile (the constant overhead per job) is amortized. The trade-off is more CI minutes total since each shard pays Julia startup separately. We can knock that down later with `julia-actions/cache` hits.

## Context

This is the "phase 1" of the testing parallelism discussion (cf. the Julia community thread on `ReTestItems.jl` / `ParallelTestRunner.jl`):

- **Why not ReTestItems?** It requires test files to contain ONLY `@testitem` / `@testsetup` blocks. Our 335 testitems live inline in `src/**/*.jl` alongside regular code, which fails its discovery (`Test files must only include @testitem and @testsetup calls`). Migrating would mean relocating every testitem out of src/ — a big refactor that's worth doing eventually but not gated on this CI win.
- **Why not ParallelTestRunner?** Parallelizes at the file level (one Julia process per file in `test/`). With our inline pattern there's only one logical "file" (`test/runtests.jl`), so file-level parallelism gives us nothing.
- **What we did instead.** Sharding via `@run_package_tests filter=...` — uses TestItemRunner's existing filter hook to pick the subset for each shard. Zero new dependencies.

## Follow-ups (out of scope)

- [ ] Migrate `@testitem` blocks from `src/` into `test/**/*_test.jl` and adopt ReTestItems for in-process testitem-level parallelism (jar's point in the slack thread: `ntestfiles < ncores < ntestitems` makes this strictly better than file-level).
- [ ] Per-shard coverage merging (codecov should already handle multi-upload).

## Test plan

- [ ] CI runs all 4 shards green on this PR
- [ ] Sum of testitems across shards equals the unsharded count (sanity check before relying on this)
- [ ] Local `Pkg.test("Piccolo")` still runs everything

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - shard ${{ matrix.shard }}/4 - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -28,8 +28,17 @@ jobs:
           - ubuntu-latest
         arch:
           - x64
+        shard:
+          - '1'
+          - '2'
+          - '3'
+          - '4'
     env:
       PYTHON: ""
+      # Distribute @testitem blocks across `shard` parallel CI jobs.
+      # Filter logic lives in test/runtests.jl.
+      TEST_SHARD: ${{ matrix.shard }}
+      TOTAL_SHARDS: '4'
     steps:
       - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v3
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,16 @@
 using TestItemRunner
 
-@run_package_tests
+# CI matrix shards the testitems across N jobs that run in parallel.
+# Locally / when unset, run everything (TOTAL_SHARDS=1).
+const TEST_SHARD    = parse(Int, get(ENV, "TEST_SHARD",    "1"))
+const TOTAL_SHARDS  = parse(Int, get(ENV, "TOTAL_SHARDS",  "1"))
+
+if TOTAL_SHARDS == 1
+    @run_package_tests
+else
+    # Stable assignment by (filename, name) hash. Same Julia version + same
+    # testitem set ⇒ same shard, so coverage is identical across runs in a
+    # single CI invocation.
+    @run_package_tests filter = ti ->
+        mod(hash((ti.filename, ti.name)), TOTAL_SHARDS) == TEST_SHARD - 1
+end


jack-champagne:
Parking this PR. Recording the next-step design before closing so it's preserved when we come back to it.

## What this PR actually buys

Verified from the live run logs (job `25938121698`):

- Sharding works correctly. All 12 jobs (3 Julia versions × 4 shards) pass.
- Wallclock improvement varies sharply by Julia version:
  - 1.10: **−44%** (~30 min → ~17 min)
  - 1.11: **−44%** (~30 min → ~17 min)
  - 1.12: **−7%** (~27 min → ~25 min)

## Why 1.12 is weak

`julia-actions/cache@v3` auto-generates its cache key from matrix context — including the shard value. Confirmed from a real cache key on this PR:

```
julia-cache;workflow=CI;job=test;os=Linux;version=1.12;os=ubuntu-latest;arch=x64;shard=1;run_id=...
```

The `shard=1` segment means **each shard has its own private cache**. On a fresh branch / Manifest change, the package gets precompiled 4× (once per shard) instead of once. On 1.12, precompile cost is large enough relative to test work that distributing test work across 4 shards is mostly cancelled out by the 4× precompile tax.

## Proposed fix: warmup-job pattern

Run a precompile job per Julia version that writes a shard-agnostic cache, then have shards restore it:

```yaml
jobs:
  precompile:
    name: Warm precompile cache (${{ matrix.version }})
    runs-on: ubuntu-latest
    strategy:
      matrix:
        version: ['1.10', '1.11', '1.12']
    steps:
      - uses: actions/checkout@v6
      - uses: julia-actions/setup-julia@v3
        with: { version: ${{ matrix.version }} }
      - uses: actions/cache@v4   # bypass julia-actions/cache@v3 to control key
        with:
          path: |
            ~/.julia/artifacts
            ~/.julia/packages
            ~/.julia/compiled
          key: julia-warm-${{ runner.os }}-${{ matrix.version }}-${{ hashFiles('**/Manifest.toml', '**/Project.toml') }}
      - uses: julia-actions/julia-buildpkg@v1
      - run: julia --project=. -e 'using Pkg; Pkg.precompile()'

  test:
    needs: precompile
    strategy:
      matrix:
        version: ['1.10','1.11','1.12']
        shard:   ['1','2','3','4']
    steps:
      - uses: actions/checkout@v6
      - uses: julia-actions/setup-julia@v3
        with: { version: ${{ matrix.version }} }
      - uses: actions/cache/restore@v4
        with:
          path: |
            ~/.julia/artifacts
            ~/.julia/packages
            ~/.julia/compiled
          key: julia-warm-${{ runner.os }}-${{ matrix.version }}-${{ hashFiles('**/Manifest.toml', '**/Project.toml') }}
      - uses: julia-actions/julia-runtest@v1
        env:
          TEST_SHARD:   '${{ matrix.shard }}'
          TOTAL_SHARDS: '4'
      # julia-processcoverage + codecov-action as in the current branch
```

Expected effect on 1.12: instead of `~25 min × 4 shards` you get `~5 min precompile warmup` + `~7 min sharded work` ≈ **~12 min wallclock**, down from ~25.

## Trade-offs

- Adds 3 sequential precompile jobs (one per Julia version) ahead of the 12 shard jobs.
- Net GitHub-Actions-minutes cost: slightly higher per run (the warmups are charged), but warmup runs once per `Manifest.toml` change, not per shard.
- Coverage path unchanged — still `julia-actions/julia-runtest@v1` in the shards, so codecov works (in contrast to the abandoned action-based path on #206).

## Why closing now

Two reasons:

1. The 1.12 net win in the current form (−7%) isn't worth the additional CI complexity by itself.
2. The warmup-pattern PR will fully replace this branch's runtests.jl + CI.yml shim — there's no value in landing this intermediate form first.

When ready to revisit: branch off `main`, implement the warmup-job pattern from the YAML above. The runtests.jl sharding filter on this branch is correct and can be reused as-is.

Related: action-based parallelism was explored in #206 (closed) — upstream issue at https://github.com/julia-actions/julia-run-testitems/issues/1.

PR #206, opened 2026-05-15
ci: parallel testitems via julia-run-testitems action (experimental)

jack-champagne:
## Summary

Swap our test step from `julia-actions/julia-runtest@v1` (sequential `Pkg.test`) to `julia-actions/julia-run-testitems@main`, which drives `TestItemRunnerCore` + `TestItemControllers` — the same in-process worker pool the VS Code Julia extension uses for parallel testitem execution.

Discovery is via `JuliaWorkspaces`, the same scanner the extension uses, so our **inline `@testitem` blocks in `src/*.jl` are picked up without any refactor**.

## Comparison vs PR #205 (sharding)

| | #205 (sharding) | this PR (workers) |
|---|---|---|
| Approach | N matrix jobs, each runs ~1/N of testitems | 1 matrix job, fan out to N workers inside one Julia process |
| Precompile cost | Paid N× (once per shard) | Paid 1× per matrix entry |
| Code changes | runtests.jl + CI.yml | CI.yml only |
| New deps | none | none in Project.toml (action installs its own) |
| Risk | low (TestItemRunner is stable) | medium (action is pinned to \`@main\`, no tags yet) |

## How it works

- \`julia-actions/install-juliaup@v2\` installs the matrix Julia version
- \`julia-actions/cache@v3\` caches the package depot (Makie / OrdinaryDiffEq etc.)
- \`julia-actions/julia-run-testitems@main\` separately caches the *runner's* depot (TestItemRunnerCore + deps), then invokes \`run_tests(pwd(); ...)\`. Workers fan out via \`Distributed.jl\` inside that single job.

## Known limitations (holding in draft)

1. **Upstream action is unreleased.** \`julia-actions/julia-run-testitems\` is labeled experimental by its maintainers and has no tagged releases — we'd be pinned to \`@main\` (or a SHA). Could break without notice. Want to wait for a tagged release before un-drafting.
2. **No equivalent of \`force_latest_compatible_version\`.** The old workflow set this to \`auto\` on 1.12 so CompatHelper / dependabot PRs actually resolved to the latest-compatible deps on 1.12 (we explicitly opted 1.10/1.11 out because we don't care as much there). The new action exposes no such input, so CompatHelper bumps would silently resolve to the existing pins and stop being exercised.

   Workaround when we revive this: add a \`Pkg.update()\` step before the testitems run, gated to 1.12:

   \`\`\`yaml
   - name: Force latest compatible deps (1.12 only)
     if: matrix.version == '1.12'
     run: julia --project=. -e 'using Pkg; Pkg.update()'
   - name: Run testitems
     uses: julia-actions/julia-run-testitems@vX  # pinned tag
     with: ...
   \`\`\`

## Status

Holding as **draft** until upstream cuts a tag. Sharding (#205) remains the fallback if we need parallelism sooner.

## Test plan
- [ ] CI green on 1.10
- [ ] CI green on 1.11
- [ ] CI green on 1.12
- [ ] Wallclock time comparable to or better than #205 across all 3 versions
- [ ] CompatHelper-style PR on 1.12 actually exercises the bumped version (after applying the workaround above)


jack-champagne:
Closing this experiment.

After validating that the architecture works (`@testmodule` + `julia-run-testitems@main` does drive in-process parallel workers via `TestItemControllers`, and tests pass), we hit two blockers for landing this in production:

1. **Coverage is hardcoded off** in the action's `main.jl` — workers don't run with `--code-coverage`, so `julia-processcoverage` only produces a zero-hit structural lcov. `codecov/patch` checks pass but only because no src/ lines were touched in this PR; real coverage data is gone.

2. **`kill_test_processes()` hangs for >1h 40m on Julia 1.10** after all 347 testitems pass. Tests finish at 02:39:42 UTC; the runner only emits `Before kill_test_processes()` and then sits idle until we cancelled. 1.11 and 1.12 tear down workers in ~4s. Without a `timeout-minutes` cap this bleeds Actions minutes.

Both blockers are upstream in `julia-actions/julia-run-testitems` / `julia-testitems/TestItemControllers.jl`. The action README labels itself as "an experiment, nothing to see at the moment" — that's accurate today.

**Path forward:** the sister PR #205 (matrix sharding) stays as the production plan — it preserves `julia-actions/julia-runtest@v1` (so coverage works) and gives most of the parallelism win without depending on an experimental stack. Once `julia-actions/julia-run-testitems` is production-ready (coverage + reliable teardown across all Julia versions), revisit.

Filing the consolidated upstream issue now; will link back here.

jack-champagne:
Upstream issue filed: https://github.com/julia-actions/julia-run-testitems/issues/1

PR #207, opened 2026-05-20
Reimplementing PiccoloOptions.bound_state (dead since 11/2025) + new PiccoloOptions.bound_state_l2 feature

gennadiryan:
### Background

`PiccoloOptions.bound_state` was introduced in QuantumCollocation.jl in June 2024 as a linear box constraint bounding each isomorphism state vector component to [-1, 1]. It was actively used in bosonic/transmon examples and defaulted to `true` for exponential integrators. The implementation was deleted in Nov 2025 during a template rewrite (QC `00af4fa`), but the struct field survived into Piccolo.jl as a dead no-op.

### Changes

**`bound_state` (restored):** Adds a `BoundsConstraint` (linear, zero NLP overhead) on each state variable component at every knot point, reproducing the original QC behavior. Applied via `apply_piccolo_options!` using the existing DirectTrajOpt `BoundsConstraint` infrastructure.

**`bound_state_l2` (new):** Adds a `NonlinearKnotPointConstraint` enforcing Re(z)^2 + Im(z)^2 <= 1 per complex component at every knot point. Tighter than the box constraint but adds Jacobian/Hessian entries. Requires awareness of the isomorphism layout:
- Ket trajectories: block layout `[Re(psi); Im(psi)]`
- Unitary trajectories: per-column interleaved `[Re(col); Im(col); ...]`
- Density trajectories: not yet supported (compact iso has non-trivial Re/Im pairing)

Both options apply only to state variables, not controls.

### Files changed

- `src/control/options.jl` — Added `bound_state_l2::Bool` field, updated docstrings
- `src/control/constraints.jl` — Added `BoundStateL2Constraint` helper with `:block` and `:interleaved_columns` layout dispatch
- `src/control/templates/_problem_templates.jl` — Added `iso_layout` kwarg to `apply_piccolo_options!`, wired both constraint branches
- `src/control/templates/smooth_pulse_problem.jl` — Updated `_apply_piccolo_options` dispatches to pass `iso_layout` per trajectory type; added `ArgumentError` for `DensityTrajectory` + `bound_state_l2`

### Test plan

- [x] `bound_state` adds `BoundsConstraint` on single and multiple state names
- [x] `bound_state` throws `ArgumentError` when `state_names` is missing
- [x] `bound_state_l2` adds `NonlinearKnotPointConstraint` with correct `g_dim` (block layout)
- [x] `bound_state_l2` throws `ArgumentError` when `state_names` is missing
- [x] `BoundStateL2Constraint` block layout: correct constraint values at known points
- [x] `BoundStateL2Constraint` interleaved_columns layout: correct for identity matrix iso-vec
- [x] `BoundStateL2Constraint` throws on invalid layout
- [x] Regression: 111 existing SmoothPulseProblem tests pass (all trajectory types)

### DIFF
--- a/src/control/constraints.jl
+++ b/src/control/constraints.jl
@@ -429,7 +429,7 @@ function BoundStateL2Constraint(
         n_pairs = n * (n - 1) ÷ 2
         function density_constraint(x)
             result = Vector{eltype(x)}(undef, n_pairs)
-            for p in 1:n_pairs
+            for p = 1:n_pairs
                 re = x[re_idx[p]]
                 im = x[im_idx[p]]
                 ρ_jj = x[dj_idx[p]]
--- a/src/control/constraints.jl
+++ b/src/control/constraints.jl
@@ -455,4 +455,76 @@ end
     @test constraint.g_dim == 1
 end
 
+@testitem "BoundStateL2Constraint block layout" begin
+    using NamedTrajectories
+    using DirectTrajOpt
+
+    N = 5
+    # 4-dim iso-vec = 2 complex components (block layout: [Re; Im])
+    traj = NamedTrajectory(
+        (ψ̃ = rand(4, N), u = rand(1, N), Δt = fill(0.1, N));
+        timestep = :Δt,
+        controls = :u,
+    )
+
+    con = BoundStateL2Constraint(:ψ̃, traj, :block)
+    @test con isa DirectTrajOpt.AbstractNonlinearConstraint
+    @test con.equality == false
+    # 2 complex components → 2 inequality constraints
+    @test con.g_dim == 2
+
+    # Evaluate constraint at a known point: ψ̃ = [0.6, 0.8, 0.0, 0.0]
+    # Complex components: z₁ = 0.6+0i, z₂ = 0.8+0i
+    # |z₁|² - 1 = -0.64, |z₂|² - 1 = -0.36 (both satisfied)
+    x = [0.6, 0.8, 0.0, 0.0]
+    g = con.g(x, nothing)
+    @test g ≈ [0.36 - 1.0, 0.64 - 1.0]
+
+    # Point violating constraint: ψ̃ = [0.8, 0.0, 0.8, 0.0]
+    # z₁ = 0.8+0.8i → |z₁|² = 1.28 > 1
+    x2 = [0.8, 0.0, 0.8, 0.0]
+    g2 = con.g(x2, nothing)
+    @test g2[1] > 0  # violated
+end
+
+@testitem "BoundStateL2Constraint interleaved_columns layout" begin
+    using NamedTrajectories
+    using DirectTrajOpt
+
+    N = 5
+    # 2×2 unitary → 8-dim iso-vec (interleaved columns: [Re(col₁); Im(col₁); Re(col₂); Im(col₂)])
+    traj = NamedTrajectory(
+        (Ũ⃗ = rand(8, N), u = rand(1, N), Δt = fill(0.1, N));
+        timestep = :Δt,
+        controls = :u,
+    )
+
+    con = BoundStateL2Constraint(:Ũ⃗, traj, :interleaved_columns)
+    @test con isa DirectTrajOpt.AbstractNonlinearConstraint
+    @test con.equality == false
+    # 2×2 = 4 complex entries → 4 inequality constraints
+    @test con.g_dim == 4
+
+    # Identity matrix I₂: iso_vec = [1, 0, 0, 0, 0, 1, 0, 0]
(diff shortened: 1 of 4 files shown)


jack-champagne (commented):
Solid restoration + thoughtful new feature, thanks @gennadiryan. Layout indexing verified against `src/quantum/primitives/isomorphisms.jl` (ket `[Re; Im]`, unitary per-column `[Re; Im]` stride-`2d`) — it matches. CI is green and codecov is 94% so I'm pretty confident to merge this one. A few notes to add in:

1. **Silent `dim` assumptions in `BoundStateL2Constraint`** (`src/control/constraints.jl:336`, `:353`). `n = dim ÷ 2` and `d = isqrt(dim ÷ 2)` truncate without checking. If a caller passes a state name whose `traj.dims[name]` violates `dim = 2d²` (unitary) or `dim` odd (ket), the constraint silently misindexes. See inline suggestions.

2. **Test coverage gaps (which are minimal)** (codecov flagged 3 lines, 94% patch coverage):
   _- No end-to-end solve demonstrating either flag actually constrains the optimizer._
   _- No test for `MultiKetTrajectory` + `bound_state_l2` (uses `:block` at `smooth_pulse_problem.jl:707`)._
   \- No test that `bound_state_l2` on `DensityTrajectory` throws the new `ArgumentError` (this is the 50%-covered line in `smooth_pulse_problem.jl`).
   \- No violation case for `:interleaved_columns` (only the identity-iso `[1,0,0,0,0,1,0,0]` is tested).

NITs:

3. **Intermediate allocations in `block_constraint`.** `NonlinearKnotPointConstraint`'s API requires `g(x)` to *return* a vector (no in-place form — `DirectTrajOpt/.../knot_point_constraint.jl:93`; the optimized path lives in `Piccolissimo.OptimizedNonlinearKnotPointConstraint`). So the return alloc is structural. But the two slice copies in `block_constraint` are eliminable with `@view` — see inline suggestion. `interleaved_constraint` already indexes scalar-by-scalar, so its only alloc is `result`, which is structural — leave it.

4. **`bound_state` is additive to existing traj-level bounds.** If a user has manually populated `traj.bounds[:ψ̃]`, `DirectTrajOpt` already emits a `BoundsConstraint` for it (`problems.jl:185`); setting `bound_state = true` adds a second. State variables aren't auto-bounded by Piccolo templates, so this is rare in practice — but worth a one-line docstring note or maybe something warning users against abuse! Open to suggestions here :)

I recommend (1) and at least the density-throw test from (2) before merge.

jack-champagne (commented):
_(folded into the review above)_

src/control/constraints.jl
```diff
@@ -312,6 +313,75 @@ function LeakageConstraint(
...
+function BoundStateL2Constraint(
+    name::Symbol,
+    traj::NamedTrajectory,
+    iso_layout::Symbol;
+    times = 1:traj.N,
+)
+    dim = traj.dims[name]
+
+    if iso_layout == :block
+        n = dim ÷ 2
```
jack-champagne, line 389: Guard against odd `dim` to avoid silent misindexing:

```suggestion
        n = dim ÷ 2
        iseven(dim) || throw(ArgumentError("block layout expects even dim; got $dim"))
```

src/control/constraints.jl
```diff
@@ -312,6 +313,75 @@ function LeakageConstraint(
...
+        function block_constraint(x)
+            re = x[1:n]
+            im = x[(n+1):2n]
```
jack-champagne, line 392: Use views to drop the two intermediate slice copies per evaluation. The broadcast result alloc is structural (NonlinearKnotPointConstraint's API requires returning a vector), but these copies are eliminable:

```suggestion
            re = @view x[1:n]
            im = @view x[(n+1):(2n)]
```

src/control/constraints.jl
```diff
@@ -312,6 +313,75 @@ function LeakageConstraint(
...
+        end
+        return NonlinearKnotPointConstraint(
+            block_constraint,
+            name,
+            traj;
+            equality = false,
+            times = times,
+        )
+    elseif iso_layout == :interleaved_columns
+        d = isqrt(dim ÷ 2)
```
jack-champagne, line 403: Same silent-misindexing guard for the unitary layout — `isqrt` rounds down if `dim` isn't `2d²`:

```suggestion
        d = isqrt(dim ÷ 2)
        dim == 2 * d^2 || throw(ArgumentError("interleaved_columns expects dim = 2d²; got $dim"))
```

jack-champagne (approved):
Approving — suggestions inline for the should-fix items from my earlier review. Test gaps and docstring nit can land in a follow-up.

PR #213, opened 2026-05-22
perf(systems): parametric eltype for H_drives — type-stable per-subst…

Rchari1:
## Summary

Adds a parametric `DD<:AbstractVector{<:AbstractDrive}` type parameter to
`QuantumSystem` and `CompositeQuantumSystem`, preventing the constructors
from widening the user's concrete drive container back to
`Vector{AbstractDrive}`. This restores type-stable dispatch on
`drive_coeff` / `drive_coeff_jac` in downstream per-substep RHS hot loops.

## Changes

- `src/quantum/systems/quantum_systems.jl`:
  - Add `DD<:AbstractVector{<:AbstractDrive}` type parameter
  - Field `H_drives::Vector{AbstractDrive}` → `H_drives::DD`
  - In matrix-input constructor: stop building `AbstractDrive[...]` for
    `linear_drives`; let Julia infer the concrete `Vector{LinearDrive}`
  - In drives-input constructor: stop wrapping with
    `collect(AbstractDrive, drives)` — preserve the caller's eltype as-is
  - In accumulator-based constructor (Pair-syntax drift/drives): add
    `drives = identity.(drives)` at the end to narrow the eltype to the
    most-specific containing Union after dynamic `push!` calls
- `src/quantum/systems/composite_quantum_systems.jl`:
  - Add matching `DD<:AbstractVector{<:AbstractDrive}` parameter
  - Field update mirrored
  - Constructor stops widening to `AbstractDrive[...]`
  - Explicit `CompositeQuantumSystem{typeof(H),typeof(G),typeof(...)}`
    construction now includes the new third parameter

## Breaking change (strict)

Downstream code that explicitly parameterized `QuantumSystem` with all 5
prior type parameters (e.g. `QuantumSystem{F1,F2,PT,DT,HD}`) needs to add
a 6th parameter. Same for `CompositeQuantumSystem` (now 3 params instead
of 2). In practice almost nobody does this — the type is used without
parameters in user code.

## Backwards-compatible

- Constructor signatures unchanged
- Field access (`sys.H_drives`) still returns the same vector content
- The `AbstractDrive`-typed accumulator constructor at top-level (e.g.,
  `AbstractDrive[]` in the function-Hamiltonian path) still works — it
  just stores its empty vector with `DD = Vector{AbstractDrive}`

## Measured downstream speedup (i=1 CNOT)

This PR is one of several stacked optimizations validated downstream in
the rqpnn workspace. Measured on CNOT i=1 K=1→4 (N=15, ketdim=10) using
Piccolissimo `feat/post-tier4-speedups-A4-A6` + this PR:

| | Pre-optimization | After full stack | Speedup |
|---|---|---|---|
| Stage 1 wall (CPU) | 180.4s | 97.8s | **1.85×** |
| Stage 2 wall (CPU, K=4 multiket) | 412.0s | 217.4s | **1.90×** |
| Total | 592.4s | 315.2s | **1.88×** |
| K=4 COHERENT fid | 0.999823 | 0.999823 (bit-equal) | — |
| Per-ket AVG | 1.000009 | 1.000009 (bit-equal) | — |

The CNOT i=1 problem is small enough that this PR's impact is
within noise of the other stack changes (dispatch overhead at 21 drives ×
~10s of µs is lost in the 100s of wall time). The structural fix is
**load-bearing for i=2 Toffoli** where:
- 21 drives × tens of dispatches per substep × N=50 knots × thousands of
  substeps gives projected 2-4× CPU RHS speedup
- Pre-fix baseline: Stage 1 = 14h CPU on the i=2 N=50 problem; this PR
  alone is projected to bring that to ~3.5-7h

## Validation

- **Bit-equal fidelity** verified on CNOT i=1 K=1→4: fid = 0.999823
  matches pre-fix baseline exactly
- **Per-ket trajectory + rollout COH** match to all printed digits
- **Precompile clean** on Julia 1.12.6 with Piccolissimo + PiccolissimoCUDAExt
  consuming the new struct

Toffoli i=2 N=50 T=100 measurement to follow in a separate report once
warm-start regeneration is complete (the saved JLD2 warm start was
generated against the pre-fix struct shape, so a fresh Stage 1 run is
required to load it under the new type signature — known JLD2 limitation
on struct parameter changes).

### DIFF
--- a/src/quantum/systems/composite_quantum_systems.jl
+++ b/src/quantum/systems/composite_quantum_systems.jl
@@ -35,14 +35,16 @@ H_coupling = 0.1 * kron(PAULIS[:Z], PAULIS[:Z])
 csys = CompositeQuantumSystem(H_coupling, [sys1, sys2], Float64[])
 ```
 """
-struct CompositeQuantumSystem{F1<:Function,F2<:Function} <: AbstractQuantumSystem
+struct CompositeQuantumSystem{F1<:Function,F2<:Function,DD<:AbstractVector{<:AbstractDrive}} <: AbstractQuantumSystem
     H::F1
     G::F2
     H_drift::SparseMatrixCSC{ComplexF64,Int}
     # Parity with QuantumSystem: drives are AbstractDrive, not raw matrices,
     # so Piccolissimo's SplineIntegrator can call dynamics_operator(::AbstractDrive)
-    # on each entry.
-    H_drives::Vector{AbstractDrive}
+    # on each entry. Parametric `DD` preserves the concrete eltype (e.g.
+    # Vector{LinearDrive} or Vector{Union{LinearDrive,NonlinearDrive}}) for
+    # type-stable iteration in hot per-substep RHS loops.
+    H_drives::DD
     drive_bounds::Vector{Tuple{Float64,Float64}}
     n_drives::Int
     levels::Int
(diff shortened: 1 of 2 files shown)


jack-champagne:
Took a closer look at the Pair-syntax constructor — flagging a small secondary issue (not a blocker for the headline win):

`drives = AbstractDrive[]` followed by `drives = identity.(drives)` causes Julia to capture `drives` as `Core.Box` inside the `H_fn`/`G_fn` closures (verified via closure field-type inspection + `@code_warntype`). This leaks into Piccolo's built-in Padé integrator path, where `Ĝ = (u_, t) -> sys.G(u_, t)` (`src/control/integrators.jl:33,42,56,65,...`) is called per substep.

Benchmark (21 drives, 10×10 sparse Hermitian): ~8% slower, +5 allocs/call. Piccolissimo's `SplineIntegrator` is unaffected since it reads `sys.H_drives` directly and the struct's `DD` parameter still captures the concrete eltype there.

**Suggestion** — rename the accumulator so `drives` is never rebound:

```julia
drives_acc = AbstractDrive[]
for d_input in H_drives_input
    ...
    push!(drives_acc, ...)
end
drives = identity.(drives_acc)   # fresh binding, no rebind
```

One-line change, no behavioral impact, free ~8% on the built-in integrator path.

PR #214, opened 2026-05-23
feat(spline): zero R_u/R_du defaults for CubicSplinePulse

aarontrowbridge:
## Summary

Default `R_u = R_du = 0` for `CubicSplinePulse`; leave `LinearSplinePulse` defaults unchanged. Users wanting smoothness regularization on cubic Hermite splines should attach `Piccolissimo.HermiteBendingEnergyRegularizer` (already on Piccolissimo `main` as of [Piccolissimo PR #83](https://github.com/harmoniqs/Piccolissimo.jl/pull/83)). Avoids a duplicate implementation in Piccolo.

### Why the defaults change

The previous `SplinePulseProblem` inherited `R_u = R_du = R = 1e-2` for **every** spline pulse type. For `CubicSplinePulse` this is actively harmful: a quadratic L2 penalty on the knot values and Hermite tangents biases the optimizer toward small-amplitude flat pulses regardless of whether that is needed.

Tested on a 2-qubit iSWAP problem:
- With `R_u = R_du = 1e-2` the optimizer stalls at F ≈ 0.91
- With `R_u = R_du = 0` it reaches F ≈ 0.98+

In both cases the pulses use < 60% of the available control bound, so this is not a bound issue — it's regularization bias.

`LinearSplinePulse` keeps its existing `R_u = R_du = R` defaults — its `du` variable is the *constrained* inter-knot slope, and the L2 penalty there is the standard smoothness term. Pass `R_u` / `R_du` explicitly to override either default.

### Why no `BendingEnergyObjective` in Piccolo

An earlier revision of this PR included a Piccolo-side `BendingEnergyObjective` (with closed-form per-segment Hessian) plus a `λ_bend` kwarg on `SplinePulseProblem`. That code has been removed in favor of `Piccolissimo.HermiteBendingEnergyRegularizer`, which:

- Already exists on Piccolissimo `main` (PR #83).
- Handles non-uniform `Δt` via per-segment timestep variables.
- Supports vector-valued `R` weighting per drive.
- Matches the documented Piccolissimo objectives pattern.

Wiring it into `SplinePulseProblem` here would require a Piccolo → Piccolissimo dependency, but Piccolissimo already depends on Piccolo — that's a cycle. The user-side attachment pattern is documented in `SplinePulseProblem`'s docstring:

```julia
using Piccolissimo
qcp = SplinePulseProblem(qtraj, N; ...)
add_objective!(qcp.prob, HermiteBendingEnergyRegularizer(qcp.prob.trajectory; R = 0.01))
```

## Changes

- `src/control/templates/spline_pulse_problem.jl`
  - `R_u::Union{Nothing, Float64, Vector{Float64}} = nothing` (was `R`); resolved to `0.0` for `CubicSplinePulse`, `R` for `LinearSplinePulse`. Same for `R_du`.
  - Same change on the `MultiKetTrajectory` method.
  - Docstring section "Default regularisation by pulse type" explaining the change, plus a "Cubic-spline smoothness regularization" section pointing at `Piccolissimo.HermiteBendingEnergyRegularizer`.

## Test plan

- [x] `SplinePulseProblem CubicSplinePulse default R_u, R_du are zero` — asserts the new zero defaults
- [x] `SplinePulseProblem LinearSplinePulse default R_u, R_du = R` — asserts the unchanged `LinearSplinePulse` defaults
- [x] `SplinePulseProblem CubicSplinePulse: defaults do not over-regularize X gate` — asserts the `QuadraticRegularizer` contribution is exactly 0.0 on a representative single-qubit X-gate problem under the new defaults
- [x] Full Piccolo test suite: 2062 passed, 0 failed (errors only from missing optional packages CairoMakie / QuantumToolbox / JET / Aqua — pre-existing environmental issue on `main`)
- [ ] CI passes


jack-champagne:
Verified that this only changes the defaults for a cubic spline pulse. Unsure if dropping regularization on the controls makes sense (`R_du` should probably be zero, but `R_u`? not so convinced). Also, there are multiple references to Piccolissimo internal objectives that should probably be removed if we want to keep them internal, or just moved into Piccolo as we determine.

jack-champagne:
Ran a 6-config ablation on a 2-qubit iSWAP using `SplineIntegrator`
(so `du` affects the dynamics) with and without
`HermiteBendingEnergyRegularizer`. Seed=214214, T=50 ns, 21 knots,
4 drives (σx⊗I, I⊗σx, σy⊗I, I⊗σy), J=2π·50 MHz.

| Config | R_u  | R_du | bend | infid    | peak \|u\| | max \|du\| | avg HF |
|---|---|---|---|---|---|---|---|
| A          | 1e-2 | 1e-2 | –    | 7.0e-08  | 0.53 | ~0    | 1.2e-1 |
| B (this PR)| 0    | 0    | –    | 1.1e-06  | 0.63 | 4.5   | 4.0e-1 |
| C          | 1e-2 | 0    | –    | 1.8e-06  | 0.09 | 4.0   | 5.4e-1 |
| D          | 1e-2 | 0    | 1e-2 | 1.5e-08  | 0.27 | ~0.1  | 6.6e-3 |
| E          | 1e-4 | 0    | 1e-2 | 3.6e-15  | 0.29 | ~0.1  | 1.9e-3 |
| F          | 0    | 0    | 1e-2 | 5.1e-15  | 0.29 | ~0.1  | 1.7e-3 |

With `SplineIntegrator` all six configs reach ≥ 99.9999% fidelity.
In the realistic operating mode (bending energy attached — D/E/F),
*lower* `R_u` gives *better* fidelity AND *better* spectrum:
F (`R_u = R_du = 0 + bend = 1e-2`) reaches infid = 5e-15 vs D's 1.5e-8.
The PR's `R_u = R_du = 0` default is the right choice when bending
energy is attached as documented.

Couldn't reproduce the F=0.91 stall on this setup — happy to re-run
in your exact iSWAP regime if you share drives / T / knots /
integrator choice.

jack-champagne:
fig_s1_pulses
fig_s2_summary
fig_s3_tangents

PR #215, opened 2026-05-23
Add pulse animation helper

dominyyin-sudo:
Addresses #59 (parameter-sweep animation; the single-pulse reveal + populations panel land in #240).

## Summary
- Adds `animate_pulse(pulses)` for parameter-sweep / optimization-snapshot pulse animations.
- Renders each frame via the existing `plot_pulse!` API (type-aware) instead of re-deriving sampling/rendering.
- Documents the sweep workflow in the visualization guide and API reference.

## Notes
- Supports `:stacked` and `:overlay` layouts.
- Reuses `animate_figure`, so `:inline` and `:record` modes behave consistently with existing animations.

## Validation
- Maintainer-verified locally (Julia 1.12 + CairoMakie): Gaussian amplitude sweep + ZOH snapshots, both layouts, render correctly. CI green (Formatter, Docs, Julia 1.10/1.11/1.12).

### DIFF
--- a/docs/literate/guides/visualization.jl
+++ b/docs/literate/guides/visualization.jl
@@ -87,6 +87,30 @@ optimized_traj = get_trajectory(qcp)
 optimized_pulse = ZeroOrderPulse(optimized_traj)
 fig = plot_pulse(optimized_pulse; title = "Optimized Pulse")
 
+# ### Pulse animation
+#
+# `animate_pulse` turns a vector of pulses into a parameter sweep animation.
+# This is useful when comparing candidate pulses from a scan or showing how a
+# waveform responds to parameter changes. Use `mode = :record` with `CairoMakie`
+# for documentation-friendly `.gif` or `.mp4` output:
+
+amplitudes = collect(range(0.2, 1.0, length = 24))
+pulse_sweep = [
+    GaussianPulse([amp, 0.5 * amp], 1.0, T; center = T / 2)
+    for amp in amplitudes
+]
+
+animate_pulse(
+    pulse_sweep;
+    mode = :record,
+    filename = "pulse_parameter_sweep.gif",
+    fps = 12,
+    parameter_values = amplitudes,
+    parameter_label = "amplitude",
+    title = "Gaussian pulse sweep",
+)
+nothing # hide
+
 # ## Basic Trajectory Plotting
 #
 # The `plot` function from NamedTrajectories.jl plots trajectory components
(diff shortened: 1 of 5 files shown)


jack-champagne:
Thanks for this — the parameter sweep is a genuinely useful addition. A couple of restructuring notes before another pass:

**1. Separate the two animations.** The single-pulse left-to-right "drawing" mode largely duplicates `animate_name(traj, :u)` (`ext/PiccoloMakieExt.jl`), which already reveals a control progressively. The new value here is the **sweep** (`animate_pulse(pulses)`) — I'd focus the PR on that. (For reference, #59's other in-scope item — showing the quantum state/population evolving *alongside* the pulse — isn't covered yet; that's the piece that would make this non-overlapping, if you want to tackle it.)

**2. Reuse the pulse plotting API.** Instead of re-deriving sampling, colors, labels, and step-vs-line logic, lean on `plot_pulse!(ax, pulse; ...)` (`src/visualizations/quantum_objects/pulse_plots.jl`). It already dispatches type-aware rendering (stairs for `ZeroOrderPulse`, lines for analytic/spline) and theming onto an existing `Axis`. A sweep frame becomes roughly `empty!(ax); plot_pulse!(ax, pulses[i])`, which lets you drop `_draw_pulse_series!`, `_animation_drive_color`, `_pulse_labels`, `_pulse_y_limits`, and the stacked/overlay duplication.

**Heads-up on the docs:** the build runs with `execute=false` (`docs/make.jl`), so the literate code isn't executed and the embedded `.gif`s are never generated — the images will 404. Either commit pre-rendered gifs as assets or drop the inline image references.

dominyyin-sudo:
Thanks for the clear direction. I pushed a follow-up commit (`b95621a`) that refocuses this PR on the parameter sweep case.

Changes made:
- Removed the single-pulse left-to-right `animate_pulse(pulse)` overload from the PR scope.
- Kept `animate_pulse(pulses)` for parameter sweeps only.
- Reworked the sweep animation to redraw each frame with the existing `plot_pulse!` API instead of duplicating pulse sampling/rendering logic.
- Removed the extra helper paths for manual step/line drawing and dropped the new inline GIF image references from the docs to avoid 404s under `execute=false`.

Validation:
- `git diff --check` passes.
- I could not run the Julia smoke test in this local environment because `julia` is not available on PATH, so I have not claimed runtime verification beyond the static check.

dominyyin-sudo:
Follow-up: I rebased the branch on the current `main` and force-pushed the conflict-clearing version. The current head is `965a5cc`.

Validation remains: `git diff --check` passes locally. I still do not have Julia available in this environment, so I have not run the Makie smoke test locally.

PR #218, opened 2026-05-26
fix(drives): isolate NonlinearDrive validator from global RNG

Rchari1:
## Summary

The `NonlinearDrive` Jacobian/Hessian validators (`validate_drive_jacobian`, `validate_drive_hessian`) sampled random control vectors via `randn(u_dim)` against the **global** `Random` stream. This means `QuantumSystem(...)` construction silently advances the user's RNG state as a side effect — and the *amount* of advance depends on `n_drives × n_samples × u_dim`.

Switch both validators to a local `MersenneTwister(0)` so construction is deterministic w.r.t. the global stream. Validation semantics are unchanged.

## Why this matters

Downstream code that calls `QuantumSystem(...)` after `Random.seed!(...)` and then samples a random initial pulse will silently land on a *different* initialization depending on which version of the validator ran. On our CNOT i=1 K=1→K=4 reproducer this caused Stage 2 truth-coherent fidelity to flip from **1.000000** to **0.027** with no other change — the post-construction `rand(...)` for the initial pulse landed in a different basin because the validator's `randn` count had drifted (207 → 252 draws between two implementations of the validator).

After this fix, `QuantumSystem(...)` is RNG-neutral: a downstream `Random.seed!(42)` actually controls the next user `rand`/`randn` deterministically.

## What changed

- **`src/quantum/systems/drives.jl`:** both validators construct `rng = Random.MersenneTwister(0)` and call `randn(rng, u_dim)`. Docstring notes added for both validators explaining the local-RNG behavior.
- **`src/quantum/systems/_quantum_systems.jl`:** `using Random`.
- **Regression test** (`@testitem "validators do not advance global RNG state"`): snapshots `randn(8)` after `Random.seed!(0)`, runs both validators, snapshots again, asserts equality. Locks the guarantee in CI so a future change re-introducing the side effect fails immediately.

No public API change. No change to validator detection power — same samples-per-drive, same `atol`.

## Test plan

- [x] **New regression test** passes locally: `validators do not advance global RNG state` → `expected == got: true`.
- [x] **Downstream reproducer** verified: running `test_cnot_i1_K1to4_nopolish_newalt.jl` (CNOT i=1 K=1→K=4 curriculum) on Piccolo `main` HEAD `f369bb6` with this patch applied produces **Stage 2 COHERENT: 1.000000**, vs **0.027** without the patch. No other Piccolo changes are required.
- [x] **JuliaFormatter v2.3.2** applied — `format(".")` reports clean repo-wide.
- [ ] CI: existing `validate_drive_jacobian` / `validate_drive_hessian` testitems still pass (they don't depend on global RNG state).

## Note

This addresses the RNG side effect only. The separate per-drive `u_dim` correctness work on `fix/nonlinear-drive-validator-globals` (commit `8ef439e`, handling `active_controls` indices beyond `n_controls`) is independent and not included here — happy to fold that in or stack a follow-up PR depending on preference.


jack-champagne:
Closing as a duplicate of #217, which is already merged on `main` with the same `MersenneTwister(0)` fix in both validators.

The two changes shipped concurrently (~60s apart) — no fault in your work, just a race. The substantive deltas in this PR were:

1. **Docstring notes on local-RNG behavior** — porting these to `main` as a small follow-up PR (link to follow).
2. **The autoformat commit** — already covered.

The downstream `active_controls`/`n_controls` fix on `fix/nonlinear-drive-validator-globals` (`8ef439e` + `2ac7741`) is the real value-add not yet on `main` — please open that as its own PR.

PR #233, opened 2026-06-05
Two qubit gate synthesis tutorial

### LINKED ISSUE #198: Add literate tutorial: 2-qubit gate synthesis with three pulse parameterizations + QuantumToolbox validation
## Preamble 
This issue is part of unitaryHACK26. You have to be [registered](https://unitaryhack.dev/register/) to complete this issue. Learn more about the Pull Request process [here](https://unitaryhack.dev/hacker-guide/) and about unitaryHACK rules [here](https://unitary.design/rules/)! A note about AI Slop: while we are open to collaboration with LLMs for unitaryHACK, fully AI-generated PRs are not acceptable. It is at the maintainers' discretion whether or not LLM-generated PRs are the right fit for the issues, and those that appear fully AI-generated may be immediately rejected. Read unitaryHACK's full AI Policy [here](https://unitaryhack.dev/ai-guide/).

## Background

Piccolo's existing literate tutorials ([`first_gate.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/first_gate.jl), [`multilevel_transmon.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/multilevel_transmon.jl), [`state_transfer.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/state_transfer.jl), [`robust_control.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/robust_control.jl)) all cover **single-qubit** examples. There's no published 2-qubit walkthrough, and no docs artifact that demonstrates **independent cross-library validation** — i.e., "Piccolo reports fidelity X; confirm it by rolling the same pulse out in [QuantumToolbox.jl](https://github.com/qutip/QuantumToolbox.jl) and getting the same answer."

A worked 2-qubit example is the single most-requested onboarding artifact, and independent-simulator confirmation is the kind of evidence that converts "curious-skim-of-the-docs" into "actually-running-it-locally." This bounty bundles both into one literate page that doubles as a regression canary — any future refactor that breaks parity with QuantumToolbox will fail the doc build.

## Minimal Reproducible Example

After this issue is resolved, the page should walk through something like:

```julia
using Piccolo, QuantumToolbox, CairoMakie

# 2-qubit transmon system, CNOT target
sys = MultiTransmonSystem(ωs, δs, gs; levels_per_transmon=3, drive_bounds=0.2)
U_goal = EmbeddedOperator(:CX, sys)

# Solve three times — one per pulse parameterization — using cached_solve!
prob_zoh    = SmoothPulseProblem(UnitaryTrajectory(sys, ZeroOrderPulse(...),    U_goal), N; Q=100.0, R=1e-2)
prob_linear = SplinePulseProblem(UnitaryTrajectory(sys, LinearSplinePulse(...), U_goal), N; Q=100.0)
prob_cubic  = SplinePulseProblem(UnitaryTrajectory(sys, CubicSplinePulse(...),  U_goal), N; Q=100.0)

Piccolo.cached_solve!(prob_zoh,    "two_qubit_zoh";           max_iter=200, verbose=false)
Piccolo.cached_solve!(prob_linear, "two_qubit_linear_spline"; max_iter=200, verbose=false)
Piccolo.cached_solve!(prob_cubic,  "two_qubit_cubic_spline";  max_iter=200, verbose=false)

# For each solved pulse, independently roll out in QuantumToolbox and
# compute U_implemented; compare gate fidelity vs Piccolo's reported value
F_piccolo, F_qutip = piccolo_vs_qutip_fidelity(prob_zoh, sys, U_goal)
@assert abs(F_piccolo - F_qutip) ≤ 1e-4   # validation gate
```

## What to implement

Add a new literate page `docs/literate/two_qubit_gate_validation.jl` (or `docs/literate/guides/two_qubit_gate_validation.jl` if you prefer a `guides/` subdir, mirroring `first_gate.jl`).

The page should:

1. **Set up a 2-qubit system** — recommend two coupled transmons via `MultiTransmonSystem` if a template exists, else hand-built drift + drives — with a target gate. **Recommended: CNOT**, or iSWAP if you want something flashier.
2. **Solve the gate three times**, once with each pulse parameterization:
   - `ZeroOrderPulse` via `SmoothPulseProblem`
   - `LinearSplinePulse` via `SplinePulseProblem`
   - `CubicSplinePulse` via `SplinePulseProblem`
3. **Validate each result with an independent QuantumToolbox rollout** — `sesolve` from the four computational basis states, reconstruct the implemented unitary, compute gate fidelity vs the target.
4. **Plot**:
   - Three pulse waveforms side by side (one panel per parameterization).
   - Population traces for one representative input state under each pulse.
   - A small comparison table: iteration count, wall time, Piccolo fidelity, QuantumToolbox fidelity, |Δ| between the two.
5. **Sidebar**: add the new page to [`docs/make.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/make.jl).

### Caching — read before writing the page

Literate pages run at every doc build. **Re-solving three optimization problems from scratch each CI run is unacceptable.** Use the existing caching machinery:

- Use [`Piccolo.cached_solve!`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/docs_cache.jl) — never bare `solve!`. Reference call sites: [`docs/literate/quickstart.jl:82`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/quickstart.jl), [`docs/literate/first_gate.jl:119`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/first_gate.jl), [`docs/literate/state_transfer.jl:56`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/state_transfer.jl).
- Use **three distinct cache names**, one per pulse parameterization — they share problem geometry but have different decision variables, so they need separate cache files: `"two_qubit_zoh"`, `"two_qubit_linear_spline"`, `"two_qubit_cubic_spline"`.
- Pick `max_iter` per pulse type so each reaches **≥ 0.999 fidelity** but the page builds in under ~30s on cache-hit. Existing 1-qubit pages use 20–100; a 2-qubit gate will likely need ~100–200. Tune once, then commit the resulting `.jld2` cache artifacts so the first cold doc build doesn't fire a fresh solve.
- During development, pass `force=true` to regenerate after editing the problem; **remove before opening the PR**.
- The QuantumToolbox rollout itself doesn't need caching — `sesolve` on a 2-qubit system with a fixed pulse is fast — but it should run *after* `cached_solve!` so it always validates against the cached pulse, never a stale one.
- `docs/make.jl` already enables `mask_cached_solve = true` at line 111 so cache hit/miss banners don't pollute the rendered page. Don't change this.

### QuantumToolbox rollout helper

[`PiccoloQuantumToolboxExt`](https://github.com/harmoniqs/Piccolo.jl/blob/main/ext/PiccoloQuantumToolboxExt.jl) currently only ships `plot_bloch` / `plot_wigner` overloads — there's **no rollout helper yet**. Hand-roll a small one for this page:

```julia
using QuantumToolbox

function rollout_with_qutip(system, pulse_traj, ψ0, tlist)
    H_drift = QuantumToolbox.Qobj(system.H_drift)
    H_drives = [QuantumToolbox.Qobj(H) for H in system.H_drives]
    drive_funcs = [t -> pulse_value(pulse_traj, i, t) for i in eachindex(H_drives)]
    H_total = [H_drift, [[H, f] for (H, f) in zip(H_drives, drive_funcs)]...]
    return sesolve(H_total, QuantumToolbox.Qobj(ψ0), tlist).states
end
```

The main technical challenge is extracting a callable `pulse_value(traj, i, t)` from a Piccolo trajectory across all three pulse types. Look at how [`extract_pulse.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/quantum/trajectories/extract_pulse.jl) exposes pulse data and how `ZeroOrderPulse` / `LinearSplinePulse` / `CubicSplinePulse` in [`primitives/pulses.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/quantum/primitives/pulses.jl) implement sampling.

## Definition of done

- New literate page renders cleanly via `Documenter.jl`.
- All three pulse parameterizations reach **≥ 0.999** gate fidelity on the test system.
- Piccolo-reported fidelity and QuantumToolbox rollout fidelity agree to **≤ 1e-4** for all three pulses (validation passes — the page asserts this).
- Three pulse-shape plots + three population/Bloch-trajectory plots embedded.
- Comparison table embedded (iter count, wall time, Piccolo fidelity, QuantumToolbox fidelity, |Δ|).
- Three `.jld2` cache files committed under wherever `cached_solve!` writes by default.
- Sidebar entry added in `docs/make.jl`.

## Stretch (optional, encouraged)

If you find yourself writing a clean `rollout_with_qutip(system, pulse, ψ0)` helper, **promote it into [`PiccoloQuantumToolboxExt`](https://github.com/harmoniqs/Piccolo.jl/blob/main/ext/PiccoloQuantumToolboxExt.jl)**. This would let future tutorials call it in one line and would naturally pair with the existing `plot_bloch` / `plot_wigner` overloads. If pursued: docstring + a small unit test in the extension's test file.

## Out of scope

- Adding or modifying problem types.
- Benchmarking solve times competitively — this is a tutorial, not a benchmark.
- A 1-qubit version (already covered by existing tutorials).
- `BangBangPulseProblem` — keeping this to spline / smooth-family parameterizations for a clean apples-to-apples comparison.

## Why this matters

"How do I do CNOT?" is the most-asked onboarding question. A worked 2-qubit gate with independent QuantumToolbox validation answers it once and gives users a copy-paste starting point — *and* serves as a regression canary that catches any future refactor that breaks integrator/`sesolve` parity. It also seeds the first tutorial coverage of the QuantumToolbox extension.

## Getting started

- [`docs/literate/first_gate.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/literate/first_gate.jl) — single-qubit version of the same idea; mirror its prose / structure.
- [`src/docs_cache.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/docs_cache.jl) — `cached_solve!` (exported at the umbrella level).
- [`docs/make.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/docs/make.jl) — docs build config (note `mask_cached_solve = true` at line 111).
- [`ext/PiccoloQuantumToolboxExt.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/ext/PiccoloQuantumToolboxExt.jl) — extension to extend if pursuing the stretch.
- [`src/quantum/trajectories/extract_pulse.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/quantum/trajectories/extract_pulse.jl) — pulse extraction utilities.
- [`src/quantum/primitives/pulses.jl`](https://github.com/harmoniqs/Piccolo.jl/blob/main/src/quantum/primitives/pulses.jl) — pulse type definitions (for sampling at arbitrary `t`).

leespen1:
## What?

Closes #198 .

Wrote and added the requested literate tutorial for 2-qubit gate synthesis.

Chose to synthesize a CNOT gate for two transmon qubits, and tried to match the style of the single-qubit tutorials that are already in the documentation.

## How?

Used `MultiTransmonSystem` to construct the system. We only model two levels per transmon, and make the coupling strength artificially high, in order to make the problem computationally cheaper so that the optimizations (when not reading cached results) are faster.

Used optimized `ZeroOrderPulse` to create a warm start pulse for `LinearSplinePulse`, which is then optimized and used to create a warm start pulse for `CubicSplinePulse`.

Piccolo fidelities are computed using Piccolo's `fidelity` function on the optimized trajectories.

QuantumToolbox fidelities are computed using `sesolve` and a short handmade fidelity function (for getting the fidelity between two unitaries).

### Optimized Pulses 
image
### Population Graphs
image

### Performance of NLP Optimization
When not reading cached results, the optimizations take several minutes, even when only modeling two levels per transmon and using an artificially high coupling strength. Each optimization took several minutes:

image

I do not see any obvious performance footguns I may have commited, and I follow the typical usage of the package showcased in the other tutorials to the point where I don't see a reason why the performance should be substandard. Is this just a normal consequence of the larger system size than usual, since doubling the size of the system doubles the number of decision variables in the NLP (which is quite large, since it scales with the number of timesteps). 

For that reason, I fear that using the states as decision variables will not scale well beyond very small (single qubit) systems, so it would be nice to get some guidance from more experienced users on whether this kind of performance is normal.

### Getting Wall Time and Number of Iterations
A `QuantumControlProblem` does not appear to expose the elapsed wall-time or number of iterations used in the IPOPT optimization, so I had to use REGEX on the cached text output of the optimization to obtain these numbers. This is very hacky and will probably confuse most people who read the tutorial. I have included a note in the tutorial saying this as well.

Is there an easier way to get the elapsed wall-time? We could also just omit these numbers from the table, but the issue #198 specifically requested that these numbers be included in a table.

If the current solution is acceptable, simple delete the TODO note.

### QuantumToolbox Extension Helper
The helper functions in this tutorial used to interface with QuantumToolbox could probably easily be moved to `PiccoloQuantumToolboxExt`, but I thought it would be best to get feedback on the implementation in the tutorial before making changes to the package source code.

## AI DIsclosure
Agentic AI was used to explore the API for Piccolo and as an assistant to help write much of the code. The source code is not "vibe-coded", all AI-generated code has been read and checked by hand, and I take full responsibility for the quality of the code. The markdown descriptions (the "literate" part of the literate tutorial) are written almost entirely by hand, with AI only being used to help with some formatting and to brainstorm.


leespen1:
I went ahead and moved the QuantumToolbox rollout interface to `PiccoloQuantumToolboxExt`.

Something I noticed but which is outside the scope of this issue is that `PiccoloQuantumToolboxExt` is only available with both `QuantumToolbox` **and** `Makie` loaded. The rollout itself is not closely related to plotting, so there may be users who want to use the rollout but don't want to load `Makie`. To address this, we could divide `PiccoloQuantumToolboxExt` into two extensions, one for functions which depend only on `QuantumToolbox`, and one for the plotting functionality that depends on `Makie` and `QuantumToolbox.`

jack-champagne:
Thanks @leespen1 — this is really well done. The warm-start chaining across the three pulse types and the independent QuantumToolbox cross-check both look great, and the plots are excellent.

On your **"Getting Wall Time and Number of Iterations"** question: you're right that there's no clean accessor today. `solve!` returns `nothing` and the optimizer/MOI stats aren't stored on the problem, so recovering iterations/wall-time after the fact requires exactly the kind of log-scraping you did. I'd rather not ship that in a tutorial — it reaches into `Piccolo.DocsCache` internals, it's fragile against IPOPT output changes, and (as you noted) it makes readers think about the cache, which they shouldn't have to.

**So let's just drop the `iters` and `solve(s)` columns** from the comparison table, along with the `cache_stats` helper and the `TODO` note. Keep the columns that carry the validation story: pulse type, `N`, `F Piccolo`, `F QuTiP`, and `|Δ|` (pic−q). That's the actual point of the page — all three parameterizations reach ≥ 0.999 and agree with QuantumToolbox to ≤ 1e-4.

This does relax the part of #198 that asked for iteration count + wall time in the table — that's on me for over-specifying it, and it won't count against you for the bounty. Exposing solver stats cleanly is a genuine gap, but it's a source-level change (the plumbing exists — `IpoptOptimizerState.iter_count`, and MOI's `BarrierIterations()` / `SolveTimeSec()` on the Ipopt optimizer), so I'll track it as a separate issue rather than block this PR on it.

I'll follow up with a few smaller content notes (a couple of typos, the `Step 5` numbering, and a `sesolve`-vs-Lindblad wording fix) shortly.

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+F_zoh_qutip, F_lin_qutip, F_cub_qutip
+
+# We need a helper function to get the number of iterations and wall time.
+#
+# !!! note "TODO"
+#     I am currently grabbing the iterations and wall time by REGEXing the IPOPT
+#     output generated by the cached solve. Is there an easier way to do this?
+#     This is very hacky and user-unfriendly (the user should not have to deal
+#     with caching), but I don't see a way to get the IPOPT iterations and
+#     walltime from a `QuantumControlProblem`.
```
jack-champagne: You're right, maybe I should not have included this in the scope or direction. Let's remove all of the regexing code and be fine with having those stats unreported.

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+# `ZeroOrderPulse` than for the other pulses because the discretized dynamics
+# used in the optimization are exact for `ZeroOrderPulse`.
+
+# ## Step 6: Visualizing Population Dynamics
+#
+# Finally, we visualize the population dynamics for each pulse type, side by
+# side. The CNOT gate flips the target qubit when the control qubit is excited:
+# ``|10\rangle \to |11\rangle``. We plot the evolution of the computational-state
+# populations starting from the initial condition ``|10\rangle``, using
+# QuantumToolbox's rollout of the dynamics.
```
jack-champagne, line 365: There was no step 5! Let's renumber these things again.

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+# + \sum_{i < j} g_{ij} \left( a_i\, a_j^\dagger + a_i^\dagger\, a_j \right),
+# ```
+# and the control Hamiltonians are
+# ```math
+# H_{\textrm{control}}(t) = \sum_{i=1}^{N} \left[ u_{x,i}(t)(a_i + a_i^\dagger) + u_y(t)i(a_i^\dagger - a_i) \right]
+# ```
+# where ``g_{ij}`` determines the coupling strength between qubits ``i`` and ``j``.
+#
+# We could build the drift and control Hamiltonians ourselves and
+# contruct the quantum system directly, but we use Piccolo.jl's
```
jack-champagne:
```suggestion
# construct the quantum system directly, but we use Piccolo.jl's
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+#
+#     1. We only model two levels of each qubit. This reduces the size of
+#        the Hilbert space, which reduces the number of 
+#        [decision variables](@ref decision-variables) in the NLP optimization.
+#     2. The coupling between the qubits is artificially high, which
+#        greatly reduces the pulse duration needed to synthesize the gate with high
+#        fidelity.
+#
+#     When synthesizing a gate for real hardware, at least 3 levels should be
+#     modeled for each qubit in order to suppress leakege to excited states
```
jack-champagne:
```suggestion
#     modeled for each qubit in order to suppress leakage to excited states
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+#     outside of the computational subspace. Additionally, ``g`` will by in the
```
jack-champagne:
```suggestion
#     outside of the computational subspace. Additionally, ``g`` will be in the
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+# Our target is the CNOT gate. Piccolo already defines this gate in `GATES[:CX]`.
+# Because we only model two levels per qubit, we could use `GATES[:CX]` as our
+# gate directly, but we will use an `EmbeddedOperator` to place the gate in
+# the system's Hilbert space and record the indices of the computational
+# subspace, so that the tutorial is still applicable when `levels_per_transmon` 
+# ``\geq 2``.
+
+U_goal = EmbeddedOperator(GATES[:CX], sys)
+
+# We set an gate duration of ``T = 10`` ns. The timestep size ``\Delta t`` is
```
jack-champagne:
```suggestion
# We set a gate duration of ``T = 10`` ns. The timestep size ``\Delta t`` is
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+#
+# For these reasons, we first optimize the `ZeroOrderPulse`. Then we convert the
+# optimized pulse to the more realistic `LinearSplinePulse` and use that as a
+# warm start for another pulse optimization. Finally, we convert the result of
+# that optimization to a `CubicSplinePulse` and use that as a warm start for
+# our final pulse optimization.
+# 
+# !!! note "Smoothness of `ZeroOrderPulse`"
+#     `ZeroOrderPulse` is not smooth in the mathematical sense, but placing a
+#     penalty on the magnitude of the amplitiude change between constant regions
```
jack-champagne:
```suggestion
#     penalty on the magnitude of the amplitude change between constant regions
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+#     good. 
+
+#-
+
+F_lin_piccolo = Piccolo.fidelity(qcp_lin)
+
+# ### Cubic Spline
+#
+# Next we optimize a cubic spline control pulse, whose control parameters
+# are control the pulse's value and the slope of the tangent line at several
```
jack-champagne:
```suggestion
# are the pulse's value and the slope of the tangent line at several
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+# `LinearSplinePulse` and `CubicSplinePulse`,  it is important to validate that
+# the reported fidelities are accurate. We can do this using
+# [QuantumToolbox.jl](https://github.com/qutip/QuantumToolbox.jl).
+#
+#
+# ### Helper Functions
+#
+# Piccolo ships a `rollout_with_qutip(system, pulse, ψ0)` helper in its
+# `PiccoloQuantumToolboxExt` extension, which loads once both `QuantumToolbox`
+# and a Makie backend are in scope. Given a a system, a pulse, and an initial
```
jack-champagne:
```suggestion
# and a Makie backend are in scope. Given a system, a pulse, and an initial
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+    sub = U_goal.subspace
+    U_target = U_goal.operator[sub, sub]
+    d = length(sub)
+    return abs2(tr(U_target' * U_impl)) / d^2
+end
+
+
+# ### Compute Fidelities with Quantum Toolbox
+#
+# With our helper functions in plase, we can compute the fidelities using
```
jack-champagne:
```suggestion
# With our helper functions in place, we can compute the fidelities using
```

docs/literate/two_qubit_gate_validation.jl
```diff
@@ -0,0 +1,435 @@
...
+    @assert abs(fd - fq) ≤ 1e-4 "$name: |F_Piccolo - F_QuTiP| = $(abs(fd - fq)) exceeds 1e-4"
+end
+println("All parameterizations reach ≥ 0.999 and agree with QuantumToolbox to ≤ 1e-4.")
+
+# We see that all pulses synthesize the CNOT gate with ``\geq 99.9`` %
+# fidelity, and that the reported fidelity agrees with the rollout fidelity
+# computed with QuantumToolbox. Fidelity reported by Piccolo matches the
+# rollout fidelity reported by QuantumToolbox matches more closely for
+# `ZeroOrderPulse` than for the other pulses because the discretized dynamics
+# used in the optimization are exact for `ZeroOrderPulse`.
```
jack-champagne:
```suggestion
# computed with QuantumToolbox. The fidelities reported by Piccolo for `ZeroOrderPulse` 
# matches the rollout fidelity reported by QuantumToolbox matches more closely than 
# any of the other pulses. This is because the discretized dynamics
# used in the optimization are exact for `ZeroOrderPulse`.
```

src/quantum/dynamics.jl
```diff
@@ -106,6 +107,42 @@ Extended in quantum_trajectories module for specific trajectory types.
...
+`QobjEvo`, with each drive modulated by the corresponding component of `pulse(t)`.
+The evolution is sampled at `n_save` points over `[0, duration(pulse)]`.
+
+Defined as a stub here; the implementation is provided by the
+`PiccoloQuantumToolboxExt` extension and loaded when both `QuantumToolbox` and a
+Makie backend are present.
+
+# Arguments
+- `system::AbstractQuantumSystem`: The system whose drift/drives define the Hamiltonian.
+- `pulse::AbstractPulse`: A callable pulse; `pulse(t)` returns the a vector containing the value of each control pulse at time `t`.
```
jack-champagne:
```suggestion
- `pulse::AbstractPulse`: A callable pulse; `pulse(t)` returns a vector containing the value of each control pulse at time `t`.
```

jack-champagne (commented):
Looking really, good! Just minor nits and removals everywhere! I think this is good to merge after these!

jack-champagne:
Hey @leespen1, thank you for your contribution. Giving a look over it now and it seems almost ready!

Just a few minor nits to fix! See my review

leespen1:
Thanks for the feedback @jack-champagne ! I applied your grammar fixes, modified the grammar a bit in the QuantumToolbox comparison paragraph, and removed `iter` and `solve(s)` from the comparison table (along with the caching and regex machinery).

leespen1:
Do I "Close with comment"? Or do I wait for you to merge?

jack-champagne:
@leespen1 kicking off the workflow runs now! Should be minor, and I can fix any formatting issues if that one flags stuff. I am going to assign you to the issue as I think you've successfully close this one out! Congrats! And thanks for your contribution! :tada:

leespen1:
Awesome, thanks for setting up this bounty for the hackathon!

PR #253, opened 2026-07-03
feat(trajectories): rollout=:none fast path for MultiKetTrajectory (ALTISSIMO 1.5)

Rchari1:
## Summary

The Piccolo half of **ALTISSIMO 1.5** (GPU-resident build & rollout — harmoniqs/Piccolissimo.jl#196). Adds a **no-ODE `rollout = :none`** construction path to `MultiKetTrajectory`, so the eager ensemble-ODE rollout in the constructor (13 min at i=3, host-OOM-adjacent at i=4) can be replaced by a fast GPU rollout downstream (`Piccolissimo.gpu_rollout!`, companion PR).

### What changed
- **`rollout::Symbol = :cpu` kwarg** on both `MultiKetTrajectory` constructors. `:cpu` (default) is unchanged; `:none` fills knot states by tiling the initial kets (a readable warm-start guess) — no ODE solve.
- **`RolloutStates` solution wrapper** with a **mutable interior**: a GPU rollout overwrites `states` in place and flips `real_states`, so the trajectory's solution type parameter `S` never changes and `rollout!`'s solution-reassignment path is never exercised.
- **`PerKetStates` shims** implement the two consumer protocols (`pk(t)` for `qtraj(t)`/NamedTrajectory sampling, `pk.u[end]` for `fidelity`). They are stored as **views** into the shared `(ketdim, K, N)` array (via a computed `_KetStatesView` type — a `Matrix` field would silently convert-and-copy, leaving shims stale after an in-place refresh).
- **Guards:** `rollout!` on a `RolloutStates` trajectory is a hard error (both methods — the reassignment would convert-error *after* the ODE ran); `fidelity` on a not-yet-refreshed placeholder warns once (a tiled guess must never be reported as a physical fidelity). Placeholder *reads* succeed (the bootstrap's `NamedTrajectory(qt,N)` sampling consumes them).

### Validation
- **`rollout = :cpu` is bit-identical to `main`**: seeded i=2 baked-C ctor, `max|Δ| = 0.0` on sampled + terminal states (two-phase main-vs-branch check).
- **`rollout = :none` unit tests** (inline `@testitem`): tiled reads, view-staleness guard (in-place mutation visible through shims), both `rollout!` errors, the `fidelity` placeholder warning, kwarg validation.
- End-to-end (via the companion Piccolissimo PR, real A100): construction **905 s → 66 s** at i=3; GPU-rollout warm-start fid matches the CPU 1e-4 reference to 6 digits.

## Test plan
- [ ] CI passes (incl. the new inline testitems)

Part of ALTISSIMO 1.5 (harmoniqs/Piccolissimo.jl#196). The Piccolissimo consumer PR (`gpu_rollout!` + device-aware sync) depends on this — **merge this first**, then drop its temporary `[sources]` pin to this branch.

### DIFF
--- a/src/quantum/trajectories/abstract_trajectory.jl
+++ b/src/quantum/trajectories/abstract_trajectory.jl
@@ -19,6 +19,7 @@ abstract type AbstractQuantumTrajectory{P<:AbstractPulse} end
 
 export AbstractQuantumTrajectory
 export UnitaryTrajectory, KetTrajectory, MultiKetTrajectory, DensityTrajectory
+export RolloutStates
 export MultiDensityTrajectory, SamplingTrajectory
 export state_name, state_names, drive_name, time_name, timestep_name
 export get_system, get_pulse, get_initial, get_goal, get_solution
--- a/src/quantum/trajectories/ensemble_trajectory.jl
+++ b/src/quantum/trajectories/ensemble_trajectory.jl
@@ -30,6 +30,62 @@ mutable struct MultiKetTrajectory{P<:AbstractPulse,S} <: AbstractQuantumTrajecto
 end
 
 """
+    PerKetStates
+
+Per-ket protocol shim for a `RolloutStates` solution. Implements the two protocols the
+`MultiKetTrajectory` solution consumers rely on:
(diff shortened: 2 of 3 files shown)


jeonghun-jj-lee:
lgtm

PR #260, opened 2026-07-25
Pulse-lifecycle hazard cleanup: detect optimizer-vs-rollout divergence, φ-aware fidelity, honest integrator_type

aarontrowbridge:
Wave 1 of the pulse-lifecycle hazard cleanup. Four changes, one behaviour change (φ, called out
below). Full suite green: **2430 passed, 0 failed, 6 broken, 25m05s.**

## 1. `rollout_divergence(qcp)` + an automatic warning from `sync_trajectory!`

The optimizer minimizes against the **collocation** solution; `fidelity(qcp.qtraj)` reports the
**ODE re-rollout**. Both numbers were already in hand at the end of `sync_trajectory!` and nothing
compared them — `inspect` even displayed them side by side without subtracting.

Implemented as a comparison of terminal **states** in iso-vector space rather than of two
fidelities: it needs no goal, no `EmbeddedOperator` and no phase convention, and it is strictly
stronger, since two different states can share a fidelity. New `check_divergence` /
`divergence_rtol` kwargs on `sync_trajectory!` and `solve!`; threshold in
`ROLLOUT_DIVERGENCE_RTOL[]`.

**Where the 5e-3 default comes from** (measured; 3-level X gate, objective/system/goal/grid/seed
held fixed, only the pairing varied):

| pairing | divergence |
|---|---|
| spline pulse + a spline integrator (correct) | 3.5e-7 … 6.4e-4 |
| spline pulse + a piecewise-constant integrator | 5.9e-2 … 1.9e-1 |

The default sits near the geometric centre of that ~90× gap. Two things this rules out: the effect
is **not cubic-only** (a linear ramp is not piecewise constant either), and it does **not shrink
with grid refinement** — it grows (5.9e-2 → 1.9e-1 as N goes 11 → 51), because a finer grid lets
the optimizer exploit the collocation model harder. More knots is not a mitigation.

⚠️ **Known coverage gap:** returns `nothing` — meaning *not applicable*, never *no divergence* — for
`MultiKetTrajectory` / `MultiDensityTrajectory`, whose `solution.u` is indexed by sub-trajectory
rather than by time, so `solution.u[end]` is not a terminal state.

**It found 16 real divergences in this repo's own suite, 8 of them above 0.5 relative** — passing
tests whose asserted fidelities are computed on a waveform the pulse does not realize.

## 2. `fidelity(qcp)` now applies the problem's stored free phases ⚠️ behaviour change

It forwarded to `fidelity(qcp.qtraj)` with no phase handling, so every `free_phase = true` problem
silently reported its **fixed-phase** fidelity — a number its objective never optimized. It now
reads the `φ_*` globals and forwards them. An explicit `phases =` still wins, so the old value
stays reachable deliberately via `phases = zeros(n)`. When `φ_*` exists but holds no data it
**errors** rather than falling back — that state is corrupt input, and guessing is the bug.

Scale, from the pulse catalog: `transmon-X-v1` stores `fidelity = 0.99999998` (free-phase) and
`fidelity_fixed_phase = 0.99980764` — as infidelity, **2.3e-8 vs 1.9e-4, four orders of magnitude.**

**No test in the suite combined `free_phase = true` with `fidelity(qcp)`**, which is why this
survived. Two regression testitems added.

`fidelity(::KetTrajectory)` also gained `phases` / `subsystem_levels`: free-phase *ket* problems are
supported but that method had no `phases` kwarg at all, making φ-awareness impossible for them. It
uses the **number-operator** convention (new exported `number_operator_phase_diag`), deliberately
not the qubit/binary decomposition that `EmbeddedOperator` goals use — the two agree for qubits and
differ above two levels.

## 3. `verify(qcp) -> (; F_optimizer, F_rollout, Δ, phases)`

Promotes a check that had been copy-pasted into demo scripts. New
`src/control/verification.jl`, included **after** `objectives.jl` because scoring the optimizer's
own collocation state needs the `*_fidelity_loss` functions defined there. Both sides get the same
phase treatment. Complements `rollout_divergence` rather than duplicating it: `verify` compares
fidelities and so exposes "the optimizer thinks it converged" directly, while `rollout_divergence`
compares states and so catches cases where two different states share a fidelity.

## 4. `integrator_type` now names the backend it returns

It accepted `:spline` and silently returned the **piecewise-constant** `BilinearIntegrator`, and
`:ensemble` called an `EnsembleSplineIntegrator` defined nowhere in Piccolo or any declared dep
(so it could only ever throw `UndefVarError` — the reason `test/jet.jl` ran `broken = true`).

Default is now `:pwc`, **behaviour-identical** to the old `:spline` default; `:spline`, `:ensemble`
and unknown values raise errors that explain and point at `Piccolissimo.SplineIntegrator`.
Deliberately **not** making the default throw: `:spline` *was* the default value, so that would
have deleted `SplinePulseProblem`'s MultiKet functionality for every caller passing nothing.
`parallel_backend` is now inert (its only consumer was the deleted branch) and warns, matching the
existing inert-`PiccoloOptions` precedent.

## 5. Duration idiom corrected at 30 sites

`sum(get_timesteps(traj))` overshoots the end time by exactly one **padded** final timestep
(`Δt = [Δt_diff; Δt_diff[end]]` at construction) — **+10% at the common N = 11**. Replaced with
`get_duration(traj) == get_times(traj)[end]`, which sums the N−1 real intervals and is correct for
fixed *and* minimum-time-variable Δt: 12 `docs/literate/` sites, 2 docstring, 16 testitem. The 31st
occurrence is the `inspect.jl` comment that already warns about the idiom. A prose warning
explaining *why* was added to the saving/loading guide, since replacing the idiom without the
reason invites its return.

## `test/jet.jl` stays `broken = true`, with the reason corrected

Its comment claimed the single remaining JET finding was `EnsembleSplineIntegrator`. Deleting that
call cleared it — and revealed **six more**. JET reports **7 findings, none of them
`EnsembleSplineIntegrator`**; all 7 are now enumerated in the comment so nobody re-derives them.
Findings 4–5 (`unembed` on a non-`EmbeddedOperator` goal inside `fidelity(::UnitaryTrajectory)`) and
7 (`_phased_ket_goal` on the free-phase display path) sit on fidelity-reporting code and are worth
doing first. So "flip `broken = true`" was **not** a valid acceptance test for this cleanup.

## Housekeeping

`Logging` added to `[extras]` / `[targets]` / `[compat]` (Aqua requires the compat bound) so
`@test_logs min_level` works in testitems.

The pulse/integrator **pairing** regression test is deliberately **not** here — it needs both a
spline and a PWC integrator, and this package ships only the latter, which would make
`BilinearIntegrator` the object of study. It lives in Piccolissimo, and a comment here records the
measured table and points at it.


aarontrowbridge:
## ⚠️ Downstream ordering: two producers must land with this, one of them regresses without it

Auditing the demos turned up **103 `fidelity(qcp)` call sites**, and the two that write artifacts are
affected in **opposite** directions by the φ change. Flagging before merge because one is a silent
regression.

### `transmon-demo` — this PR **breaks** it

```julia
fid_fixed = fidelity(qcp)              # relied on fidelity(qcp) being φ-BLIND
fid_free  = free_phase_fidelity(qcp)
```

Once `fidelity(qcp)` is φ-aware, `fid_fixed` silently becomes the **free-phase** number and the two
stored fields collapse to the same value. Nothing errors; the artifact just quietly loses the
distinction it exists to record.

Fixed in place (transmon-demo is not a git repo, so there is no PR to link): `fid_fixed` now requests
`fidelity(qcp; phases = zeros(n))` explicitly, φ is persisted to `global_vars`, and
`fidelity_convention` is recorded.

### `atoms-demo` — this PR **fixes** it → [atoms-demo#12](https://github.com/harmoniqs/atoms-demo/pull/12)

It saved φ next to the *fixed-phase* fidelity, so the stored number never matched the stored φ. φ
awareness makes them consistent. That PR also writes both conventions explicitly and clears the
+1·Δt duration bias at 42 sites.

### Catalog data → [armonissima#43](https://github.com/harmoniqs/armonissima/pull/43)

Repairs the existing entries. Of 9 `free_phase = true` entries, only **1** still has recoverable φ
(`rydberg-CZ-quera-v1`, in its JLD2 `global_vars` — the earlier audit checked `metadata.toml` only
and concluded none did). One (`rydberg-CZ-v1`) has neither φ nor a labelled convention and is
genuinely unreproducible.

### Suggested merge order

1. **atoms-demo#12** and the transmon-demo edit (both are correct before *and* after this PR)
2. **this PR**
3. **armonissima#43** (independent — pure metadata)

Merging this PR first is safe for atoms-demo but leaves transmon-demo writing two identical fields
until its fix lands. Since that edit is already applied locally, the practical risk is only to
someone running transmon-demo from a stale checkout.

### Also worth knowing

`transmon-demo/scripts/ingest_to_catalog.jl` turned out to be **dead**: it targeted
`<demos>/../amico/catalog/pulses`, i.e. `~/.amico/ops/catalog/pulses` via the compat symlink, a path
that no longer exists — combined with `mkpath`, it would silently create an empty directory and write
entries nobody reads. That explains its schema drift (it wrote `duration_ns`; every real entry uses
`duration_us`, unconverted, so it disagreed with its siblings by 1000×). It now resolves the
armonissima catalog explicitly, **errors** rather than inventing a destination, derives `pulse_type`
and `N_knots` from the loaded pulse instead of hardcoding them, and warns that the mount is
read-only so changes go via PR.

PR #262, opened 2026-07-28
fix: thread per-state weights through all coherent ket fidelity surfaces

### LINKED ISSUE #261: Coherent ket objectives silently discard per-state weights
> [!IMPORTANT]
> ## Problem
> A user who supplies per-state weights to a multi-ket trajectory and then builds a problem with the coherent objective gets **no weighting and no error**. The ensemble objective helper takes `weights` as a positional argument and uses it only in the incoherent branch; the coherent branch drops it and builds the unweighted objective. Weighted coherent fidelity is a meaningful quantity — the downstream carrier constructor supports it — so this is a silent semantic loss, not an unsupported configuration.
>
> ## Approach
> Thread the weights through the coherent branch: generalize the coherent fidelity from an unweighted mean of overlaps to a weighted one normalized by the weight sum, and pass the weights from the ensemble helper. Unweighted callers are unaffected because uniform weights recover the current formula exactly.
>
> ## Approaches Considered
> - **Thread the weights through (chosen)** — smallest change, matches what callers already believe is happening.
> - **Error when weights are supplied with the coherent objective** — rejected: the quantity is well-defined and already supported downstream, so refusing it removes capability to fix a bug.
> - **Document the drop and leave it** — rejected: the argument is accepted, so the documentation would be describing a trap rather than removing one.
>
> ## Scope
> **In:** weighted coherent fidelity; passing weights through the ensemble objective helper in every template that calls it; a test that distinct weights change the objective value.
> **Out:** any change to how derivatives are supplied for the resulting objective — that is a downstream concern, tracked separately.
>
> ## Assumptions / Open Qs
> - **Assumed:** normalization by the weight sum, so uniform weights reproduce today's value bit-for-bit.
> - **Open:** whether weights should be normalized once at construction or applied at evaluation. Construction is preferable if it keeps the loss's captured data self-describing.

---

## Acceptance Criteria

1. Building a coherent objective from a trajectory carrying non-uniform weights produces an objective whose value depends on those weights.
2. Uniform weights reproduce the current unweighted value exactly, so existing results do not move.
3. Every template that calls the ensemble objective helper passes weights through on the coherent branch, not only the incoherent one.
4. A test asserts that two different weight vectors give two different objective values for the same trajectory — the assertion that would have caught this.

## Key Decisions

- Weighted coherent fidelity normalizes by the weight sum, so it is a weighted mean of overlaps rather than a weighted sum, and uniform weights are the identity case.
- The weights become part of the loss closure's captured data, which keeps the objective self-describing for anything downstream that inspects it.

## Notes

Found while auditing objective derivative paths for harmoniqs/Piccolissimo.jl#321. Worth coordinating: downstream recognition of coherent objectives currently reconstructs an *unweighted* factor, so a genuinely weighted loss would stop being recognized unless that recognition also reads the captured weights. Tracked on the Piccolissimo side at harmoniqs/Piccolissimo.jl#322 — the two changes are complementary, and landing this one alone would cost weighted-coherent users a fast path.

### LINKED ISSUE #263: Free-phase coherent objective and coherent fidelity constraints still discard per-state weights
> [!IMPORTANT]
> ## Problem
> #261 made the *fixed-phase* coherent ket objective weight-aware, but three sibling surfaces still compute an unweighted coherent fidelity and accept no weights at all: the free-phase coherent ket objective, and both methods of the final coherent ket fidelity constraint. The result is worse than a uniform gap — the same pulse-problem template now means two different things depending on an unrelated `free_phase` flag, silently honoring per-state weights on one branch and discarding them on the other.
>
> ## Approach
> Thread weights through the remaining surfaces exactly as #261 did, reusing the weight-canonicalization helper that #261 introduced so every coherent surface agrees on the definition by construction rather than by three independent re-derivations.
>
> ## Approaches Considered
> - **Thread the weights through (chosen)** — the fidelity kernel already takes weights after #261, so this is plumbing plus tests, and it removes a live divergence inside one template.
> - **Error when weights are combined with the free-phase branch** — rejected: the quantity is well-defined, and erroring would break configurations that work today (weights are currently ignored, not rejected).
> - **Leave it and document the asymmetry** — rejected: documenting that a flag silently changes whether your weights apply describes a trap instead of removing one.
>
> ## Scope
> **In:** per-state weights on the free-phase coherent ket objective; on both final-coherent-ket-fidelity constraint methods; the three pulse-problem templates passing the trajectory's weights on their free-phase branch; tests that distinct weights move the value on each surface.
> **Out:** the definition of weighted coherent fidelity — settled in #261, this issue only propagates it. Downstream derivative/fast-path recognition of weighted losses — that is Piccolissimo's side, tracked at harmoniqs/Piccolissimo.jl#322.
>
> ## Assumptions / Open Qs
> - **Assumed:** the #261 normalization carries over unchanged — weighted mean of overlaps over the weight sum, with uniform weights the bit-for-bit identity case.
> - **Assumed:** weights apply to the *phased* overlaps, i.e. weighting composes with the phase rotation rather than reordering it.
> - **Open:** whether a template that builds both an objective and a fidelity constraint should feed them the same weights automatically, or keep them independently specified. No template builds the constraint today, so this only becomes a question if one starts to.

---

## Acceptance Criteria

1. A free-phase coherent ket objective built from a trajectory carrying non-uniform weights produces a value that depends on those weights.
2. Both final-coherent-ket-fidelity constraint methods (fixed-phase and free-phase) accept per-state weights, and their residual depends on them.
3. Every pulse-problem template passes the trajectory's weights on its free-phase branch, so the two branches of one template agree on whether weights apply.
4. Omitted or uniform weights reproduce today's values exactly on all three surfaces — existing results do not move.
5. For each of the three surfaces, a test asserts that two different weight vectors give two different values — the assertion that would have caught this.

## Key Decisions

- Reuse the weight-canonicalization helper added in #261 (validate, collapse uniform to the unweighted path, otherwise normalize once at construction) rather than re-deriving normalization per surface. One definition, three call sites.
- Uniform and omitted weights must route to the *unweighted* code path, not merely to a weighted path that happens to be mathematically equivalent — that is what makes criterion 4 hold bit-for-bit and keeps the closure shape downstream recognition matches on.

## Constraints & Invariants

- Do not change the meaning of weights on the incoherent branch, where they scale each state's `Q`. That composition is deliberate and separate from the coherent mean.
- The constraint and objective forms of coherent fidelity must not drift apart: a user pairing a weighted objective with a weighted constraint should be constraining the same quantity the objective minimizes.

<details><summary>Prior Art / Patterns</summary>

Follow PR #262 (the #261 fix) as the pattern end-to-end: it added the `weights` keyword to the fixed-phase coherent ket objective, introduced the canonicalization helper alongside the fidelity kernel in the quantum-objectives module, and threaded weights through the shared ensemble-objective helper that the smooth, spline, and bang-bang templates all call.

The surfaces to change here are the free-phase coherent ket objective in that same objectives module, both constructors of the final coherent ket fidelity constraint in the constraints module (exported user-facing API, not currently built by any template), and the free-phase branch of the three pulse-problem templates — each already binds the trajectory's weights locally for the non-free-phase branch, so the value is in scope at the call site.

Testing pattern from #262: build deterministic states with deliberately asymmetric overlaps (e.g. one overlap 1, another ½) so a weighted mean is analytically predictable, assert the exact expected value for two weight vectors, and assert the uniform case with `===` rather than `≈`.

</details>

## Source

Follow-up to #261 (fixed-phase coherent objective), implemented in PR #262 — this issue covers the surfaces that PR explicitly left out of scope. Related: harmoniqs/Piccolissimo.jl#322 (downstream recognition of weighted coherent losses).

## Notes

Found while implementing #261: the templates already compute the trajectory's weights before branching on `free_phase`, and pass them on one branch only. That makes this a small, well-localized change — the reason it is a separate issue is scope discipline on #261, not difficulty.

jeonghun-jj-lee:
Closes #261 · Closes #263

Weighted coherent ket fidelity, threaded through **every** coherent surface in one PR. (Originally split as #262 + #264; collapsed at the author's request — #264 is closed in favour of this.)

## Problem

The ensemble objective helper accepted `weights` but used them only on the *incoherent* branch. The coherent branch built an unweighted objective, so a user who supplied per-state weights to a multi-ket trajectory and built a problem with the coherent objective got **no weighting and no error** — a silent semantic loss, since weighted coherent fidelity is well-defined and already supported downstream.

Auditing the fix surfaced that this was one of **four** surfaces with the same drop:

| Surface | Symptom before |
|---|---|
| Coherent ket objective (fixed-phase) | weights accepted by the helper, discarded |
| Coherent ket objective (free-phase) | no `weights` parameter at all |
| `FinalCoherentKetFidelityConstraint` (fixed-phase) | no `weights` parameter at all |
| `FinalCoherentKetFidelityConstraint` (free-phase) | no `weights` parameter at all |

The templates made it worse than a uniform gap: they bind `weights = qtraj.weights` *before* branching on `free_phase`, then passed them on one branch only — so an unrelated flag silently decided whether your weights applied.

Each was reproduced end-to-end before fixing (identical objective values for opposite weightings):

| Path | Value for `[0.8, 0.1, 0.1]` **and** `[0.1, 0.1, 0.8]` |
|---|---|
| `SmoothPulseProblem`, coherent | `93.50805551940275` |
| `SmoothPulseProblem`, free-phase | `93.40922440120845` |
| `SplinePulseProblem`, free-phase | `99.89899382966321` |
| `BangBangPulseProblem`, free-phase | `93.49036291965412` |

## Change

- `coherent_ket_fidelity(ψ̃s, goals; weights)` — the mean of overlaps becomes a **weighted** mean normalized by the weight sum: `F = |∑ᵢ wᵢ⟨ψᵢ_goal|ψᵢ⟩ / ∑ᵢ wᵢ|²`. Only weight ratios matter.
- `coherent_fidelity_weights` — one canonicalization helper (validate, collapse uniform, normalize once at construction), reused by all four surfaces so they agree on the definition by construction rather than by four independent re-derivations.
- `weights` keyword on both coherent objectives and both constraint methods.
- `_ensemble_ket_objective`'s coherent branch and all three templates' free-phase branch pass the trajectory's weights.

In the free-phase forms, weights apply to the **phased** overlaps, so weighting composes with the phase rotation rather than reordering it.

## Existing results do not move

Uniform and omitted weights are routed to the *unweighted code path*, so values are **bit-for-bit** what they were before weights existed — not merely equal up to rounding. Asserted with `===`, including `fill(1/3, 3)` where `1/n` is not representable.

This also matters for [Piccolissimo.jl#322](https://github.com/harmoniqs/Piccolissimo.jl/issues/322): downstream recognition of coherent objectives reconstructs an *unweighted* factor. Because unweighted and uniform-weighted callers keep the exact closure shape they had, they keep their fast path; only genuinely non-uniform weights produce a loss that #322 needs to teach recognition to read.

## Tests

Every assertion was verified **RED against the unfixed code before being made green** — including reverting each one-line template thread individually to confirm the end-to-end tests actually catch the drop, rather than passing for unrelated reasons.

| Acceptance criterion | Covered by |
|---|---|
| #261 — objective value depends on trajectory weights | `SmoothPulseProblem honors MultiKetTrajectory weights (coherent)` |
| #261 — uniform reproduces today's value exactly | `===` assertions across all surfaces |
| #261 — every template threads on the coherent branch | `_ensemble_ket_objective coherent kwarg` (all three templates share the helper) |
| #263 — free-phase objective honors weights | `CoherentKetFreePhaseInfidelityObjective` + smooth/spline/bang-bang end-to-end |
| #263 — both constraint methods honor weights | `FinalCoherentKetFidelityConstraint honors per-state weights` |
| both — two weight vectors → two values | asserted at kernel, objective, constraint, and template level |

`FinalCoherentKetFidelityConstraint` had **no test coverage at all** before this. Its new test also asserts the residual equals `final_fidelity` minus the fidelity a weighted objective drives to one, so the objective and constraint forms cannot drift apart.

## Two things outside the issues' scope, called out deliberately

**1. `size_threshold` raised in `docs/make.jl`.** The auto-generated API reference `lib.md` rendered to 500.82 KiB against a 500 KiB limit — over by 840 bytes — failing the Documentation job. It grows monotonically with every docstring in the package, so the old value sat on a knife's edge where any docstring addition trips it (the comment already read `# 400 KiB` next to a 500 KiB value, i.e. it had been bumped once before without updating the comment). Raised to 1 MiB for real headroom; other pages keep the guard. Required to get this PR green.

**2. The bang-bang free-phase test is objective-wiring only.** Coupling phase globals into bang-bang's dynamics needs `HermitianExponentialIntegrator`, which lives downstream in Piccolissimo, so the default-integrator guard rejects `free_phase` outright and Piccolo's own suite cannot build a valid free-phase bang-bang problem. Passing an explicit integrator bypasses the guard, which is enough to reach the objective-construction branch — but the resulting problem's dynamics ignore θ, so the test asserts wiring and never solves. Documented inline so the limitation is not mistaken for coverage of the dynamics.


jack-champagne:
Merged `main` into this branch to clear the conflict that appeared when #260 landed (`a353bdbc`). One real conflict, plus one semantic thing worth your attention that the merge does **not** address.

## The conflict: `docs/make.jl`

You and #260 independently raised Documenter's `size_threshold`, for the same reason — `lib.md` was sitting exactly at its 500 KiB limit, so any new docstring anywhere broke the docs build. Different measurements and different values:

- #260 measured 511.4 KiB → raised to 700 KiB
- this PR measured 500.82 KiB → raised to 1 MiB

Resolved to **1 MiB**, since both PRs *add* docstrings and the merged page is larger than either measured alone; 700 KiB would be another knife's edge. Kept #260's comment detail (Documenter's own default is 200 KiB, and the trailing "400 KiB" comment on the old value was stale). The durable fix is still splitting `lib.md` per-module.

`src/control/templates/spline_pulse_problem.jl` auto-merged cleanly. Loads fine after the merge.

## Worth a look: `verify` will report unweighted numbers

Not introduced by the merge, and not a blocker — flagging it because it is invisible.

This PR makes `coherent_ket_fidelity` weighted: `|∑ᵢ wᵢ⟨ψᵢ_goal|ψᵢ⟩ / ∑ᵢ wᵢ|²`. But `fidelity(::MultiKetTrajectory)` lives in `src/quantum/trajectories/rollouts_extensions.jl`, which this PR doesn't touch, and #260's `_multiket_optimizer_side_fidelity` in `verification.jl` hardcodes the unweighted `abs2(overlap / n)`.

So both sides of `verify(qcp)` stay unweighted and remain consistent *with each other* — meaning `verify` reports ΔF ≈ 0, a confident agreement, on two numbers neither of which is what the objective actually minimized for a weighted MultiKet problem.

That's the same shape as the φ bug #260 just fixed (fidelity reporting a number the objective never optimized), one axis over: weights instead of phases. It won't surface as a conflict, a test failure, or a divergence warning. Consistent with #263 already being open about the remaining weight plumbing.

Sequencing note: #252 rewrites `verification.jl` (`verify` now returns a `VerificationReport`, `Δ` renamed to `ΔF`), so a weight-aware optimizer side is best written after that lands rather than against the current NamedTuple.

🤖 pushed by Jack's session — happy to revert the merge commit if you'd rather rebase it yourself

PR #264, opened 2026-07-28
fix: thread per-state weights through remaining coherent ket surfaces

### LINKED ISSUE #263: Free-phase coherent objective and coherent fidelity constraints still discard per-state weights

jeonghun-jj-lee:
Closes #263 · **Depends on #262** — this PR is stacked on `#262`'s branch and targets it as base, because it builds on the `weights` keyword and canonicalization helper that #262 introduces. **Retarget to `main` once #262 merges.**

## Problem

#261/#262 made the *fixed-phase* coherent ket objective weight-aware. Three sibling surfaces still computed an unweighted coherent fidelity and accepted no weights:

- the free-phase coherent ket objective
- `FinalCoherentKetFidelityConstraint`, fixed-phase method
- `FinalCoherentKetFidelityConstraint`, free-phase method

The templates were the visible symptom. They bind `weights = qtraj.weights` *before* branching on `free_phase`, then passed them on one branch only — so after #262, the same template silently meant two different things depending on an unrelated flag. Reproduced end-to-end before the fix: `SmoothPulseProblem` with `free_phase=true` returned `93.40922440120845` for both `[0.8, 0.1, 0.1]` and `[0.1, 0.1, 0.8]`; `SplinePulseProblem` likewise returned `99.89899382966321` for both.

## Change

- `CoherentKetFreePhaseInfidelityObjective(...; weights)` — weights apply to the **phased** overlaps, so weighting composes with the phase rotation rather than reordering it.
- `FinalCoherentKetFidelityConstraint(...; weights)` on both methods.
- Smooth, spline, and bang-bang templates pass the trajectory's weights on their free-phase branch, matching their fixed-phase branch.

All four surfaces reuse `coherent_fidelity_weights` from #262 rather than re-deriving normalization, so they agree on the definition by construction. Omitted and uniform weights route to the unweighted path — bit-for-bit unchanged, asserted with `===`.

## Tests

| Acceptance criterion | Covered by |
|---|---|
| 1. Free-phase objective value depends on trajectory weights | `CoherentKetFreePhaseInfidelityObjective` (extended) + `SmoothPulseProblem …free_phase=true` end-to-end |
| 2. Both constraint methods accept weights; residual depends on them | `FinalCoherentKetFidelityConstraint honors per-state weights` (new) |
| 3. Every template threads weights on the free-phase branch | smooth + spline end-to-end; see bang-bang note below |
| 4. Omitted/uniform weights reproduce today's values exactly | `===` assertions on all four surfaces |
| 5. Two weight vectors → two values, per surface | asserted at objective, constraint (both methods), and template level |

Each new assertion was verified RED against the unfixed code before being made green — including reverting the one-line template thread to confirm the smooth and spline end-to-end tests actually catch the drop.

`FinalCoherentKetFidelityConstraint` had **no test coverage at all** before this. Its new test also asserts the constraint residual equals `final_fidelity` minus the fidelity a weighted objective drives to one, so the objective and constraint forms cannot drift apart.

## Notes

**Bang-bang is changed but not tested end-to-end.** Its MultiKet free-phase path throws (`requires a custom integrator that supports global variables`) before the objective is built, so it is unreachable with the default integrator. The one-line change is identical to the other two templates and the shared objective constructor is directly tested; building a custom integrator purely to exercise it was disproportionate. Flagging rather than silently claiming coverage.

**Local verification:** the full suite is 2274 pass / 0 genuine failures under `--project=.`; the 56 errors there are all missing test-only deps (CairoMakie, QuantumToolbox, JET, Aqua) that live in `test/Project.toml` and are present in CI.

### DIFF
--- a/src/control/constraints.jl
+++ b/src/control/constraints.jl
@@ -1,7 +1,11 @@
 module QuantumConstraints
 
 using ..QuantumObjectives
-using ..QuantumObjectives: ket_fidelity_loss, unitary_fidelity_loss, coherent_ket_fidelity
+using ..QuantumObjectives:
+    ket_fidelity_loss,
+    unitary_fidelity_loss,
+    coherent_ket_fidelity,
+    coherent_fidelity_weights
 
 using DirectTrajOpt
 using LinearAlgebra
@@ -85,9 +89,9 @@ end
 
 Create a final fidelity constraint using coherent ket fidelity across multiple states.
 
-Coherent fidelity: F = |1/n ∑ᵢ ⟨ψᵢ_goal|ψᵢ⟩|²
+Coherent fidelity: F = |∑ᵢ wᵢ ⟨ψᵢ_goal|ψᵢ⟩ / ∑ᵢ wᵢ|²
 
-This constraint enforces that all state overlaps have aligned phases, which is 
+This constraint enforces that all state overlaps have aligned phases, which is
 essential when implementing a gate via multiple state transfers (e.g., MultiKetTrajectory).
 
 # Arguments
@@ -95,6 +99,12 @@ essential when implementing a gate via multiple state transfers (e.g., MultiKetT
 - `ψ̃_names::Vector{Symbol}`: Names of isomorphic state variables in trajectory
 - `final_fidelity::Float64`: Minimum fidelity threshold (constraint: F ≥ final_fidelity)
 - `traj::NamedTrajectory`: The trajectory
+
+# Keyword Arguments
+- `weights::Union{Nothing, AbstractVector{<:Real}}=nothing`: Per-state weights on the
+  coherent mean of overlaps. Pass the same weights as the objective so the constrained
+  quantity is the one being minimized. `nothing` or uniform weights give the unweighted
+  fidelity |1/n ∑ᵢ ⟨ψᵢ_goal|ψᵢ⟩|².
 
 # Example
 ```julia
@@ -108,13 +118,18 @@ function FinalCoherentKetFidelityConstraint(
     ψ_goals::Vector{<:AbstractVector{<:Complex}},
     ψ̃_names::Vector{Symbol},
     final_fidelity::Float64,
-    traj::NamedTrajectory,
+    traj::NamedTrajectory;
+    weights::Union{Nothing,AbstractVector{<:Real}} = nothing,
 )
     n_states = length(ψ_goals)
     @assert length(ψ̃_names) == n_states "Number of names must match number of goals"
 
     # Convert goals to ComplexF64
     goals = [ComplexF64.(g) for g in ψ_goals]
+
+    # Normalize once at construction, so the constraint captures self-describing
+    # weights and matches the fidelity a weighted objective drives to one
+    ws = coherent_fidelity_weights(weights, n_states)
 
     # Get component info for extracting states from concatenated vector
     state_dims = [traj.dims[name] for name in ψ̃_names]
@@ -129,7 +144,7 @@ function FinalCoherentKetFidelityConstraint(
         end
 
         # Constraint: final_fidelity - F_coherent ≤ 0
-        return [final_fidelity - coherent_ket_fidelity(ψ̃s, goals)]
+        return [final_fidelity - coherent_ket_fidelity(ψ̃s, goals; weights = ws)]
     end
 
     return NonlinearKnotPointConstraint(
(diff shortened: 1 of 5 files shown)


jeonghun-jj-lee:
Collapsed into #262 at the author's request — that PR now carries all six commits and closes both #261 and #263, targeting `main` directly. Closing this in favour of it; no work is lost.

The bang-bang coverage gap flagged in this PR's notes is also now closed there (`fcdf679e`).

